PR #8, opened 2020-02-22
Fixed the incorrect dimensions of the return array when bootstrapping.

dosssman:
- Changed observations back to obs
- When bootstrapping, there was a missing `returns=returns[:-1]` which was borking the script.
- Some cleanups


dosssman:
Let the experiments beginning ...

dosssman:
Wow that was fast haha

vwxyzjn:
:)

PR #12, opened 2020-03-02
This adds value loss clipping and Advantage normalization to ppo2_continuou_actions.

dosssman:
Greetings.

A little bit late, but this is an attempt to incorporate the value loss clipping and the advantage normalization as optimization triclks to the PPO for continuous action.

Whlle they might not yield significant results yet (in the light of a few experiments), they do seem to work:
[Advantage Norm Effect](https://app.wandb.ai/dosssman/cleanrl.ppo_norm_adv/reports/Investigating-the-effect-of-Adv-Norm--Vmlldzo2MTA5NQ/edit)
[Value loss clipping effect](https://app.wandb.ai/dosssman/cleanrl.ppo_clip_vloss/reports/Clipped-V-Loss-vs-Standard-MSE--Vmlldzo2MTA4Nw/edit)


dosssman:
I shall next try out the weights rollback on KL breaching while running more experiments regarding the tricks above.

PR #13, opened 2020-03-12
Added layer norm to policy network and entropy bonus to policy loss.

dosssman:
Glad if I could have your input on the entropy computation and how it is added to the policy loss.

Sorry for the # DEBUG lines, but I have to run right now.

### DIFF
--- a/cleanrl/ppo2_continuous_action.py
+++ b/cleanrl/ppo2_continuous_action.py
@@ -147,7 +147,6 @@ if __name__ == "__main__":
                         help='Toggles wheter or not to use a clipped loss for the value function, as per the paper.')
     parser.add_argument('--pol-layer-norm', action='store_true', default=False,
                        help='Enables layer normalization in the policy network')
-    # TODO ? Experiment with value function layer norm too, as it gave quite good results in SAC
 
     args = parser.parse_args()
     if not args.seed:
@@ -198,10 +197,8 @@ class Policy(nn.Module):
         self.logstd = nn.Parameter(torch.zeros(1, output_shape))
 
         if args.pol_layer_norm:
-            # Layer Normalization
             self.ln1 = torch.nn.LayerNorm(120)
             self.ln2 = torch.nn.LayerNorm(84)
-            # TODO: Condisder adding a self.bn_mean for the last layer too ?
 
         if args.weights_init == "orthogonal":
             torch.nn.init.orthogonal_(self.fc1.weight)
@@ -379,7 +376,7 @@ while global_step < args.total_timesteps:
 
     # Advantage normalization
     if args.norm_adv:
-        advantages = (advantages - advantages.mean()) / advantages.std() # TODO: Correct formula btw ?
+        advantages = (advantages - advantages.mean()) / advantages.std()
 
     # Optimizaing policy network
     # First Tensorize all that is need to be so, clears up the loss computation part
--- a/cleanrl/ppo2_continuous_action.py
+++ b/cleanrl/ppo2_continuous_action.py
@@ -223,10 +223,8 @@ class Policy(nn.Module):
         x = torch.tanh(x)
 
         x = self.fc2(x)
-
         if args.pol_layer_norm:
             x = self.ln2(x)
-
         x = torch.tanh(x)
 
         action_mean = self.mean(x)
@@ -319,10 +317,6 @@ while global_step < args.total_timesteps:
         values[step] = vf.forward(obs[step:step+1])
 
         with torch.no_grad():
-            # DEBUG: Tracking the logits
-            logits = pg(obs[step:step+1])
-            logits = logits[0][0].cpu().numpy()
-
             action, logproba = pg.get_action(obs[step:step+1])
 
         actions[step] = action.data.cpu().numpy()[0]
--- a/cleanrl/ppo2_continuous_action.py
+++ b/cleanrl/ppo2_continuous_action.py
@@ -111,7 +111,7 @@ if __name__ == "__main__":
                        help='the discount factor gamma')
     parser.add_argument('--gae-lambda', type=float, default=0.97,
                        help='the lambda for the general advantage estimation')
-    parser.add_argument('--ent-coef', type=float, default=0.0, # Set to 0.0 by default to measure the actual gain. Will convene on a proper value later.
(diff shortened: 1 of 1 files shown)


cleanrl/ppo2_continuous_action.py
```diff
@@ -232,6 +258,31 @@ def get_logproba(self, x, actions):
...
+
+        # print( "# DEBUG Entropy shape")
+        # print( entropy.shape)
+
+        # print( "# DEBUG: Manual entropy computation")
+        # print( - logprobs)
+        # print( "# DEBUG: Pytorch entropy computation")
+        # print( dist.entropy().sum(1))
+
+        return entropies
```
vwxyzjn: The `get_entropies` function could be combined with
```python 
    def get_action(self, x):
        mean, logstd = self.forward(x)
        std = torch.exp(logstd)
        probs = Normal(mean, std)
        action = probs.sample()
        return action, probs.log_prob(action).sum(1)

    def get_logproba(self, x, actions):
        # Note: Converts actions to tensor
        if not isinstance(actions, torch.Tensor):
            actions = preprocess_obs_fn(actions)

        action_mean, action_logstd = self.forward(x)
        action_std = torch.exp(action_logstd)
        dist = Normal(action_mean, action_std)
        return dist.log_prob(actions).sum(1)
```

and get something like this:

```python
    def get_action(self, x, action=None):
        mean, logstd = self.forward(x)
        std = torch.exp(logstd)
        probs = Normal(mean, std)
        action = probs.sample() if not action
        return action, probs.log_prob(action).sum(1), dist.entropy().sum(1)
```

If you decided to go with Manual entropy computation, then do the entropy like in https://github.com/vwxyzjn/cleanrl/blob/master/cleanrl/ppo_continuous_action.py
```python
        newlogproba = pg.get_logproba(obs[:step+1], torch.Tensor(actions[:step+1]))
        # newvalues = vf.forward(obs[:step+1]).flatten() DO we generate a new values from the current policy?
        ratio =  torch.exp(newlogproba - torch.Tensor(logprobs[:step+1]))
        surrogate1 = ratio * torch.Tensor(advantages[:step+1])
        surrogate2 = ratio.clamp(1 - args.clip_coef, 1 + args.clip_coef) * torch.Tensor(advantages[:step+1])
        policy_loss = - torch.mean(torch.min(surrogate1, surrogate2))
        vf_loss = torch.mean((values[:step+1] - torch.Tensor(returns[:step+1])).pow(2))

        entropy_loss = torch.mean(torch.exp(newlogproba) * newlogproba)

        total_loss = policy_loss + args.vf_coef * vf_loss + args.ent_coef * entropy_loss
        optimizer.zero_grad()
        total_loss.backward(retain_graph=True)
```

Also, I don't quite see why `rsample()` is necessary here.

cleanrl/ppo2_continuous_action.py
```diff
@@ -378,7 +439,13 @@ def forward(self, x):
                                 (1.+args.clip_coef) * advantages,
                                 (1.-args.clip_coef) * advantages).to(device)
 
-        policy_loss = - torch.min(ratio * advantages, clip_adv).mean()
+        # Adding entropy bonus: This is computed using the logprobs and the rsampled actions
+        # so the gradients can also be back propagated through it (?)
+        # Does the entropy function of a distribution backpropagate ? Probably not.
```
vwxyzjn: Yes. See 
image

cleanrl/ppo2_continuous_action.py
```diff
@@ -378,7 +439,13 @@ def forward(self, x):
...
+        # Also, this method does not used the old action sampled from and old policy to get the log probs
+        # Any discussion on what would be the difference ?
```
vwxyzjn: If you are using Pytorch entropy computation, then the sampled actions has nothing to do with the entropy, which itself is calculated solely by the standard deviation of the normal distribution.

cleanrl/ppo2_continuous_action.py
dosssman: Thanks for the suggestions.

I have run some experiments with both methods for the entropy computation and while the average results are mostly similar, the manual computed entropy seems better, as per the following report : [WANDB Report PPO2 Entropy Computation](https://www.reddit.com/r/MachineLearning/comments/f5kkxb/discussion_lessons_learned_from_my_failures_in/?utm_medium=android_app&utm_source=share).

Also moved the entropy computation to the network update section.

cleanrl/ppo2_continuous_action.py
dosssman: Thanks. It makes more sense now.
Anyway, let's use the manual computation, as OpenAI Baseline do the same.

cleanrl/ppo2_continuous_action.py
dosssman: Thanks. I am so lazy, as always hahaha.
Now I know it is possible to easily instanciate a `Variable` to create distributions and such.

cleanrl/ppo2_continuous_action.py
```diff
@@ -208,8 +216,19 @@ def __init__(self):
...
+            x = self.ln1(x)
+        x = torch.tanh(x)
+
+        x = self.fc2(x)
+
+        if args.pol_layer_norm:
+            x = self.ln2(x)
+
+        x = torch.tanh(x)
+
```
vwxyzjn: Delete the spaces, please. They seem fairly random.

cleanrl/ppo2_continuous_action.py
```diff
@@ -300,6 +319,10 @@ def forward(self, x):
         values[step] = vf.forward(obs[step:step+1])
 
         with torch.no_grad():
+            # DEBUG: Tracking the logits
+            logits = pg(obs[step:step+1])
+            logits = logits[0][0].cpu().numpy()
+
```
vwxyzjn: This shouldn't be here as this is in the `cleanrl` official folder. I do think they should be added to a separate file like in `cleanrl/experiments/ppo2_continuous_action_debug.py`

cleanrl/ppo2_continuous_action.py
dosssman: My bad. I was probably "high" when working on that part hahaha

PR #15, opened 2020-04-06
PPO2 # fix value clipping

vwxyzjn:
This is related to #12 .

Relevant code can be found in https://github.com/openai/baselines/blob/ea25b9e8b234e6ee1bca43083f8f3cf974143998/baselines/ppo2/model.py#L69


vwxyzjn:
@dosssman

vwxyzjn:
Hmm, interesting. Let's stick with the original implementation for now. We should look further into this.

PR #36, opened 2020-10-16
unify replay buffer

### LINKED ISSUE #35: Both dqn_atari and dqn_atari_visual use different ReplayBuffers compared to other implementations
Thanks for creating this repo - I always have difficulty understanding RL algorithms when they're spread across huge libraries.

One question, both the `dqn_atari.py` and `dqn_atari_visual.py` both use the following `ReplayBuffer` class:
```python
# modified from https://github.com/openai/baselines/blob/master/baselines/deepq/replay_buffer.py
class ReplayBuffer(object):
    def __init__(self, size):
        self._storage = []
        self._maxsize = size
        self._next_idx = 0

    def put(self, data):
        if self._next_idx >= len(self._storage):
            self._storage.append(data)
        else:
            self._storage[self._next_idx] = data
        self._next_idx = (self._next_idx + 1) % self._maxsize

    def sample(self, batch_size):
        idxes = np.random.choice(len(self._storage), batch_size, replace=True)
        obses_t, actions, rewards, obses_tp1, dones = [], [], [], [], []
        for i in idxes:
            data = self._storage[i]
            obs_t, action, reward, obs_tp1, done = data
            obses_t.append(np.array(obs_t, copy=False))
            actions.append(np.array(action, copy=False))
            rewards.append(reward)
            obses_tp1.append(np.array(obs_tp1, copy=False))
            dones.append(done)
        return np.array(obses_t), np.array(actions), np.array(rewards), np.array(obses_tp1), np.array(dones)
```

However, `dqn.py`, `c51.py`, `c51_atari.py`, `cs51_atari_visual.py`, `sac_continuous_action.py`, and `td3_continuous_action.py` all use the following `ReplayBuffer`:

```python
# modified from https://github.com/seungeunrho/minimalRL/blob/master/dqn.py#
class ReplayBuffer():
    def __init__(self, buffer_limit):
        self.buffer = collections.deque(maxlen=buffer_limit)
    
    def put(self, transition):
        self.buffer.append(transition)
    
    def sample(self, n):
        mini_batch = random.sample(self.buffer, n)
        s_lst, a_lst, r_lst, s_prime_lst, done_mask_lst = [], [], [], [], []
        
        for transition in mini_batch:
            s, a, r, s_prime, done_mask = transition
            s_lst.append(s)
            a_lst.append(a)
            r_lst.append(r)
            s_prime_lst.append(s_prime)
            done_mask_lst.append(done_mask)

        return np.array(s_lst), np.array(a_lst), \
               np.array(r_lst), np.array(s_prime_lst), \
               np.array(done_mask_lst)
```

As far as I am aware, they look identical - but wanted to know if there was some implementation reason for why they are different?

bentrevett:
where PrioritizedReplayBuffer is not used, unify the implementation of the ReplayBuffer, closes #35

### DIFF
--- a/cleanrl/dqn_atari.py
+++ b/cleanrl/dqn_atari.py
@@ -404,32 +404,29 @@ env.observation_space.seed(args.seed)
 # respect the default timelimit
 assert isinstance(env.action_space, Discrete), "only discrete action space is supported"
 
-# modified from https://github.com/openai/baselines/blob/master/baselines/deepq/replay_buffer.py
-class ReplayBuffer(object):
-    def __init__(self, size):
-        self._storage = []
-        self._maxsize = size
-        self._next_idx = 0
+# modified from https://github.com/seungeunrho/minimalRL/blob/master/dqn.py#
+class ReplayBuffer():
+    def __init__(self, buffer_limit):
+        self.buffer = collections.deque(maxlen=buffer_limit)
+    
+    def put(self, transition):
+        self.buffer.append(transition)
+    
+    def sample(self, n):
+        mini_batch = random.sample(self.buffer, n)
+        s_lst, a_lst, r_lst, s_prime_lst, done_mask_lst = [], [], [], [], []
+        
+        for transition in mini_batch:
+            s, a, r, s_prime, done_mask = transition
+            s_lst.append(s)
+            a_lst.append(a)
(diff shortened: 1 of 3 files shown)


vwxyzjn:
Looks 👍

PR #38, opened 2021-01-07
SAC CQL for continuous tasks.

dosssman:
SAC CQL for continuous actions, following closely the original implementation.

Limited experiments can on Hopper-v2 can be found here:

https://wandb.ai/cleanrl/cleanrl.benchmark/reports/Offline-Continuous-Benchmark--VmlldzozODgxODA


cleanrl/cql_continuous/sac_cql.py
```diff
@@ -0,0 +1,396 @@
...
+
+class Policy(nn.Module):
+    def __init__(self, input_shape, output_shape, env):
+        super(Policy, self).__init__()
+        self.fc1 = nn.Linear(input_shape, 256) # Better result with slightly wider networks.
+        self.fc2 = nn.Linear(256, 256)
+        self.mean = nn.Linear(256, output_shape)
+        self.logstd = nn.Linear(256, output_shape)
+        # action rescaling
+        self.register_buffer("action_scale",
```
vwxyzjn, line 154: So what if we replace `action_scale` with just python scalar, would it be ok? Becasue `action_scale` is not really being updated, so you could do things like

`action_scale = (env.action_space.high - env.action_space.low) / 2.)`

cleanrl/cql_continuous/sac_cql.py
```diff
@@ -0,0 +1,396 @@
...
+    writer.add_scalar("charts/episode_reward", eval_stats["test_mean_return"], global_step)
+    writer.add_scalar("global_step", global_step, global_step)
+    if args.autotune:
+        writer.add_scalar("losses/alpha_loss", alpha_loss.item(), global_step)
+    if args.with_lagrange:
+        writer.add_scalar("losses/alpha_prime_loss", alpha_prime_loss.item(), global_step)
+        writer.add_scalar("losses/alpha_prime", log_alpha_prime.exp().item(), global_step)
+
+writer.close()
+env.close()
```
vwxyzjn, line 396: please add a newline in the end

cleanrl/cql_continuous/sac_cql.py
```diff
@@ -0,0 +1,396 @@
...
+            for param, target_param in zip(qf1.parameters(), qf1_target.parameters()):
+                target_param.data.copy_(args.tau * param.data + (1 - args.tau) * target_param.data)
+            for param, target_param in zip(qf2.parameters(), qf2_target.parameters()):
+                target_param.data.copy_(args.tau * param.data + (1 - args.tau) * target_param.data)
+        
+        # Stop iteration over the dataset
+        if k >= args.epoch_length-1:
+            break
+    
+    eval_stats = test_agent(env, pg)
```
vwxyzjn, line 374: since `test_agent` is not really reused, could we remove the function and simply move its content here?

cleanrl/cql_continuous/sac_cql.py
```diff
@@ -0,0 +1,396 @@
...
+    print("[E%4d|I%8d] -- PLoss: %.3f -- QLoss: %.3f -- CQLQloss: %.3f --  Train Mean Ret: %.3f" %
```
vwxyzjn, line 375: Do we really read losses from the stdout? if not, maybe just print out global step and episode reward like all other scripts. To make comparison with other algo, maybe each `global_step` could mean a gradient step.

cleanrl/cql_continuous/sac_cql.py
```diff
@@ -0,0 +1,396 @@
...
+        "test_max_return": np.max(returns),
+        "test_max_length": np.max(lengths),
+        "test_min_return": np.min(returns),
+        "test_min_length": np.min(lengths)
+    }
+
+    return eval_stats
+
+# Training Loop
+global_step = 0 # Tracks the Gradient updates
```
vwxyzjn, line 274: Maybe setup the loop in the following way instead:

```
for global_step in range(args.total_timesteps):
    if global_step % args.train_frequency == 0:
        s_obs, s_actions, s_rewards, s_next_obses, s_dones = next(data_loader)
        s_obs, s_actions, s_rewards, s_next_obses, s_dones = s_obs.to(device), s_actions.to(device), s_rewards.to(device), s_next_obses.to(device), s_dones.to(device)
        # ..... training code
    if global_step % args.eval_frequency == 0:
        #  .....  eval code
        # 
```

This will have similar structure as the CQL code.

cleanrl/cql_continuous/sac_cql.py
dosssman: I now recalled the logic behind having the action scale and the action bias as registered buffer:

First, when trying `action_scale = (env.action_space.high - env.action_space.low) / 2.)`, there is the error of multiplying a tensor by the scalar `self.action_scale`, as per the following error log:
```
python sac_cql.py 
Warning: Flow failed to import. Set the environment variable D4RL_SUPPRESS_IMPORT_ERROR=1 to suppress this message.
No module named 'flow'
Warning: CARLA failed to import. Set the environment variable D4RL_SUPPRESS_IMPORT_ERROR=1 to suppress this message.
No module named 'carla'
pybullet build time: Nov 26 2019 09:35:14
/home/d055/anaconda3/envs/cleanrl/lib/python3.6/site-packages/gym/logger.py:30: UserWarning: WARN: Box bound precision lowered by casting to float32
  warnings.warn(colorize('%s: %s'%('WARN', msg % args), 'yellow'))
/home/d055/anaconda3/envs/cleanrl/lib/python3.6/site-packages/gym/logger.py:30: UserWarning: WARN: Box bound precision lowered by casting to float32
  warnings.warn(colorize('%s: %s'%('WARN', msg % args), 'yellow'))
Traceback (most recent call last):
  File "sac_cql.py", line 281, in <module>
    next_state_actions, next_state_log_pi, _ = pg.get_action(s_next_obses)
  File "sac_cql.py", line 174, in get_action
    action = y_t * self.action_scale + self.action_bias
  File "/home/d055/anaconda3/envs/cleanrl/lib/python3.6/site-packages/torch/tensor.py", line 480, in __array__
    return self.numpy()
TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first
```

I believe this is what was addressed by having `self.action_scale` and `self.action_bias` as th.FloatTensor as in:
https://github.com/vwxyzjn/cleanrl/blob/c1fb51a7c32596ef3133af4617f92d2095051c66/cleanrl/sac_continuous_action.py#L139

However, this seems to require an overload of the `to(device, ...)` method, so that `self.action_scale` and `self.action_bias` are on the same device as the `y_t` used in `action = y_t * self.action_scale + self.action_bias`:
https://github.com/vwxyzjn/cleanrl/blob/c1fb51a7c32596ef3133af4617f92d2095051c66/cleanrl/sac_continuous_action.py#L171

The idea behind using `register_buffer` was thus to remove the need for having to declare a FloatTensor and moving it to device via overload of the `to()` method.
**Namely, any registered buffer will be automatically copied to the `device` specified when directly calling the `to()` method of the mother class.**, which I find to be a useful trick as further simplification of the code.

cleanrl/cql_continuous/sac_cql.py
dosssman: Done. Quite curious about the reason for the last line though ?

cleanrl/cql_continuous/sac_cql.py
dosssman: Indeed.

cleanrl/cql_continuous/sac_cql.py
dosssman: Mainly used for debug purpose in the early implementation phase.

Changed to `print(f"global_step={global_step}, episode_reward={avg_eval_return}")` following the other scripts
(actually, `ppo_procgen.py`).

cleanrl/cql_continuous/sac_cql.py
dosssman: - Removed the `n-epochs`  and `epoch-length` parser argument, replaced by `total-timesteps`.
- Modified the dataset class to match that in `offline_dqn_cql` (IterableDataset, etc...)
- Update the training loop to match the one in your comment, to a certain extent:

```
for global_step in range(args.total_timesteps):
    # Oneliner batch sample and moving to device ?
    s_obs, s_actions, s_rewards, s_next_obses, s_dones = [i.to(device) for i in next(data_loader)]

    # ..... training code
    
    if global_step % args.eval_frequency == 0:
        #  .....  eval code
        #  .... logging various metrics to wandb, etc...
```

cleanrl/cql_continuous/sac_cql.py
dosssman: From additional experiments, using register_buffer properly helps minimize calls to the `.to(device)` function, making the code run slightly faster.

PR #42, opened 2021-04-02
Video length increase for Procgen

bragajj:
Video length increased to 600 from prior value of 100 for ppo_procgen_fast.py and ppg_procgen_fast.py. Video length of 100 proved to be too short for truly assessing or watching the gameplay of some games in the Procgen benchmark. Additionally, pybullet import was removed from ppo_procgen_fast.py as it was an unnecessary import.


vwxyzjn (approved):
Thank you!

PR #44, opened 2021-04-26
Fix bibtex entry

justinkterry:
Your current one gives an error in latex


vwxyzjn (approved):
Thanks for the catch!

PR #48, opened 2021-05-12
Fix hyperlink to actually go to Weights and Biases

sudo-michael:
Previously, the hyperlink went to https://github.com/vwxyzjn/cleanrl/blob/master/wandb.com


vwxyzjn:
Thank you!!

Issue #53, opened 2021-09-14
Broken links on README


FelipeMartins96:
## Problem Description
Broken links on Algorithms Implemented section of README 

## Current Behavior
Links are pointing to unavailable pages

## Expected Behavior
Link should point to desired code

## Possible Solution
Remove links to removed code or change links to new location

## Steps to Reproduce
The link to the following codes on README implemented algorithms section are pointing to old locations, resulting in page 404:

experiments/ppo_self_play.py
experiments/ppo_microrts.py
experiments/ppo_simple.py
experiments/ppo_simple_continuous_action.py

not sure if there are more links, i found these one as they are on the experiments folder which does not exist anymore

vwxyzjn:
Hey @FelipeMartins96  thanks for reporting the problem. Yeah those links are broken due to some cleaning up. Would you mind making a PR removing the references to these links?

FelipeMartins96:
sure! Will do that

PR #54, opened 2021-09-16
Remove links to deleted code on README algorithms

FelipeMartins96:
#53 
Remove links to algorithms which were removed, but still mentioned on readme.


vwxyzjn:
Thank you so much!

PR #55, opened 2021-10-07
Add paper plotting utilities 

vwxyzjn:
This PR adds a bit of paper plotting utilities for CleanRL.

Try running

```bash
poetry install -E plot
poetry run python cleanrl_utils/plot.py
```

And you would see something like below:

image

### DIFF
--- a/cleanrl_utils/paper_plot.py
+++ b/cleanrl_utils/paper_plot.py
@@ -11,7 +11,7 @@ from distutils.util import strtobool
 import matplotlib as mpl
 sns.set_style("whitegrid")
 mpl.rcParams['text.usetex'] = True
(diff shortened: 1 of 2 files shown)


dosssman:
All good on my side, the script worked and generated all the plots.

Not sure if this also happens on your side, but I got a matplotlib deprecation warning in the plot.py
```
cleanrl_utils/plot.py:13: MatplotlibDeprecationWarning: Support for setting the 'text.latex.preamble' or 'pgf.preamble' rcParam to a list of strings is deprecated since 3.3 and will be removed two minor releases later; set it to a single string instead.
  mpl.rcParams['text.latex.preamble'] = [r'\usepackage{amsmath}'] #for \text command
```

I ran the code on a cleanly installed machine (Ubuntu 20), with a clean conda environment, so I had to installed system wide `texlive` and related to get the script to work.

PR #62, opened 2021-10-16
Automatically Download Atari Roms

vwxyzjn:
Basically automatically download Atari ROMs via `pip install autorom[accept-rom-license]`


dosssman:
Don't the ROMs come with gym[atari] anymore ?

Issue #57, opened 2021-10-07
Documentation Site


vwxyzjn:
## Problem Description
Although CleanRL generally has a simplistic implementation, it will be desirable to have a documentation site for some situations. For example, I'm not sure where to put instructions on how to do start and resume with CleanRL's scripts. See #33, #14.

dosssman:
Are the start and resume independent of WANDB and AWS ?

dosssman:
> > Are the start and resume independent of WANDB and AWS ?
> 
> it’s dependent on wandb because we need to store the timestep somehow.

In such case, I would say the relevant instructions could go to the wandb / cloud integration documentation section.

vwxyzjn:
I am a little in favor of mkdocs, becuase in https://fastapi.tiangolo.com/tutorial/ you can see like an animated terminal and that in the other pages you can highlight code

vwxyzjn:
Closing this issue now. Docs site is already live at https://docs.cleanrl.dev

PR #64, opened 2021-10-20
Prototype Documentation Site

vwxyzjn:
This is a prototype documentation site for #57, building on top of https://squidfunk.github.io/mkdocs-material/.

Here is the deployed site: https://docs.cleanrl.dev/

* Make Spyder as an optional dependency.


vwxyzjn:
Merging the docs branch as is. Will make modifications to it in separate PRs.

PR #65, opened 2021-10-20
Cloud Utilities Improvement 

vwxyzjn:
This PR improves the cloud workflow. See the changed readme `cloud/README.md`.

PR #67, opened 2021-10-21
Support Python 3.7.1+

vwxyzjn:


vwxyzjn:
Thanks @dosssman for testing. This will also help unblock #65: the PyTorch docker container still uses python 3.7

Issue #68, opened 2021-10-22
DDPG Actor missing 1 argument: 'env'


dosssman:
Identified while testing changes for PR #67 

https://github.com/vwxyzjn/cleanrl/blob/502f0f3abd805799d98b2d89a2564b6470b3dad0/cleanrl/ddpg_continuous_action.py#L167

When running:
```bash
python ddpg_continuous_actions.py
```

returns the error:
```
$ python ddpg_continuous_action.py 
/home/d055/anaconda3/envs/cleanrl-py3.7.1/lib/python3.7/site-packages/ale_py/roms/utils.py:90: DeprecationWarning: SelectableGroups dict interface is deprecated. Use select.
  for external in metadata.entry_points().get(self.group, []):
pybullet build time: Oct 11 2021 20:59:00
/home/d055/anaconda3/envs/cleanrl-py3.7.1/lib/python3.7/site-packages/gym/spaces/box.py:74: UserWarning: WARN: Box bound precision lowered by casting to float32
  "Box bound precision lowered by casting to {}".format(self.dtype)
Traceback (most recent call last):
  File "ddpg_continuous_action.py", line 167, in <module>
    actor = Actor().to(device)
TypeError: __init__() missing 1 required positional argument: 'env'
```

vwxyzjn:
Thanks for this catch! Would you mind submitting a PR for fixing it?

PR #69, opened 2021-10-22
ddpg_continuous: Addded env argument to actor and target actor

dosssman:
Tested.

Fixes Issue #68


vwxyzjn:
Thanks for the fix!

PR #83, opened 2021-11-04
Add PPO Atari LSTM example

vwxyzjn:
Referenced resources:

[Illustrated Guide to LSTM's and GRU's: A step by step explanation](https://www.youtube.com/watch?v=8HyCNIVRbSU)

[openai/baselines: Initialize the hidden states of LSTM to be zeros](https://github.com/openai/baselines/blob/ea25b9e8b234e6ee1bca43083f8f3cf974143998/baselines/common/models.py#L179)

[openai/baselines: Mask out observations at the end of episode](https://github.com/openai/baselines/blob/ea25b9e8b234e6ee1bca43083f8f3cf974143998/baselines/common/models.py#L141)

Takes about ~40 lines of code to implement (see [diff](https://www.diffchecker.com/dsYQsP1U) here) and the pilot run is at [here](https://wandb.ai/costa-huang/cleanRL/runs/prautp3y?workspace=user-costa-huang)

I guess the only thing is here we use LSTM and frame stack at the same time....


cleanrl/ppo_atari_lstm.py
```diff
@@ -0,0 +1,375 @@
...
+        b_dones = dones.reshape(-1)
+        b_advantages = advantages.reshape(-1)
+        b_returns = returns.reshape(-1)
+        b_values = values.reshape(-1)
+        b_lstm_states = (
+            lstm_states[0].permute((1,0,2,3)).reshape(1, -1, agent.lstm.hidden_size),
+            lstm_states[1].permute((1,0,2,3)).reshape(1, -1, agent.lstm.hidden_size)
+        )
+
+        # Optimizaing the policy and value network
```
dosssman, line 294: Just small typo haha

dosssman:
Great addition to the family.

Having framestacking enabled while using LSTM does not _feel_ very efficient, however.
Have you come across related works that discuss this aspect ?

I recall reading some papers which added RNN's into DQN or a similar algorithm that was used on pixel-based tasks, and they did mention that they removed the frame-stacking, but at the same time its not like the practice of keeping the frame-stacking is forbidden I think.

Could we further flesh this out in case you are not in a rush to get this PR merged ?
Thanks

dosssman:
Currently experimenting with `ppo_atari_lstm.py` by making two changes:
1. Use a single frame instead of frame stacking at line 102:
```python
...
env.gym.wrappers.FrameStack(env,1)
...
```
2. Adapt the CNN channel to match the single frame observation at line 121
```python
...
layer_init(nn.Conv2d(1, 32, 8, stride=4))
...
```
I am not sure how this plays with the `MaxSkipEnv` wrapper though...

The run should be located [here](https://wandb.ai/cleanrl/cleanRL/runs/2sl7y706?workspace=user-dosssman)

dosssman:
Additional runs here. Let's see how it goes:

https://wandb.ai/cleanrl/ppo_lstm_dbg/reports/PPO-Atari-LSTM-and-LSTM-No-Framestack-attempt-comparison--VmlldzoxMjA2NTAx

cleanrl/ppo_memory_env.py
```diff
@@ -0,0 +1,369 @@
...
+    args = parser.parse_args()
+    args.batch_size = int(args.num_envs * args.num_steps)
+    args.minibatch_size = int(args.batch_size // args.num_minibatches)
+    # fmt: on
+    return args
+
+
+class TestMemoryEnv(gym.Env):
+    """
+    final state  [0,0,0,1,0]    [0,0,0,0,1]
```
yooceii, line 83: First, why is the cartpole state length 5 instead of 4?
Second, why is it only integers and 5 states?
Is there any reference of this env?

cleanrl/ppo_atari_lstm.py
```diff
@@ -0,0 +1,375 @@
...
+    # ALGO Logic: Storage setup
+    obs = torch.zeros((args.num_steps, args.num_envs) + envs.single_observation_space.shape).to(device)
+    actions = torch.zeros((args.num_steps, args.num_envs) + envs.single_action_space.shape).to(device)
+    logprobs = torch.zeros((args.num_steps, args.num_envs)).to(device)
+    rewards = torch.zeros((args.num_steps, args.num_envs)).to(device)
+    dones = torch.zeros((args.num_steps, args.num_envs)).to(device)
+    values = torch.zeros((args.num_steps, args.num_envs)).to(device)
+    lstm_states = (
+        torch.zeros((args.num_steps, agent.lstm.num_layers, args.num_envs, agent.lstm.hidden_size)).to(device),
+        torch.zeros((args.num_steps, agent.lstm.num_layers, args.num_envs, agent.lstm.hidden_size)).to(device),
```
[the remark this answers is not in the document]
vwxyzjn: That’s just how LSTM kind of works in Pytorch, they require a tuple of hidden states to be passed in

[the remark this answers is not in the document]
vwxyzjn: When using GRU the implementation would have been more wlwgant

cleanrl/ppo_memory_env.py
vwxyzjn: I made up this env.

vwxyzjn:
@dosssman Thanks for adding the additional runs. I have also managed to run using the original openai/baselines code. See https://wandb.ai/cleanrl/cleanRL/reports/LSTM-no-framestack--VmlldzoxMjEwNDI3

yooceii (approved):

dosssman (approved):
Roger that.

The PPO Atari LSTM Framestack does seem to work, but underperforms when compared to either the PPO_Atari by default, or PPO_Atari_LSTM.
The latter two seems to achieve roughly similar results, at least on the Breakout task, which suggest the LSTM component is not really useful for this task at least.

Furthermore, the baselines PPO without Frame stacking also seems to have similar result to the `ppo_atari_lstm_noframestack.py`, which suggest the `ppo_atari_lstm.py` works as expected.

All good on my side.
Great work as always.

cleanrl/ppo_memory_env_lstm.py
```diff
@@ -0,0 +1,407 @@
...
+        clipfracs = []
+        for epoch in range(args.update_epochs):
+            # np.random.shuffle(b_inds)
+            np.random.shuffle(envinds)
+            for start in range(0, args.num_envs, envsperbatch):
+                end = start + envsperbatch
+                mbenvinds = envinds[start:end]
+                mb_inds = flatinds[mbenvinds].ravel()
+
+                _, newlogprob, entropy, newvalue, _ = agent.get_action_and_value(
```
cpuheater, line 340: input to the lstm during update phase should be (seq, batch , feature) but it's  ( batch , seq, feature)

dosssman: Nice catch.

vwxyzjn: Sorry I didn't quite get this. I have

```
b_lstm_states[0][:,mb_inds].shape
Out[2]: torch.Size([1, 256, 512])
```

which I assume is `(seq, batch, feature)`?

vwxyzjn:
If you see the beginning of training, the baselines LSTM actually performs best.

image

I suspect the difference could have come from two places

1. Layer initialization: the original implementation does orthogonal initialization with weights 1 and bias 0 for the LSTM stuff
([https://github.com/openai/baselines/blob/ea25b9e8b234e6ee1bca43083f8f3cf974143998/baselines/a2c/utils.py#L84-L86](https://github.com/openai/baselines/blob/ea25b9e8b234e6ee1bca43083f8f3cf974143998/baselines/a2c/utils.py#L84-L86))
2. By default, there are 128 LSTM units (https://github.com/openai/baselines/blob/ea25b9e8b234e6ee1bca43083f8f3cf974143998/baselines/common/models.py#L187), whereas in our implementation we had 512 for atari and 64 for memory env.
3. Finally, I have only applied the LSTM to the actor and not the critic, so this is an implementation error.


Also, I can confirm the hidden states don't get updated (so the original implementation does not regenerate states), as shown in the screenshot below.

image

vwxyzjn:
So the original implementation didn't really implement the dual network architecture... See https://github.com/openai/baselines/blob/ea25b9e8b234e6ee1bca43083f8f3cf974143998/baselines/common/policies.py#L165, so we will largely keep the critics in `ppo_memory_env_lstm.py` independent of memory for now.

cleanrl/ppo_memory_env_lstm.py
cpuheater: Out[2]: torch.Size([1, 256, 512]) , 1 is not a sequence but a batch and 256 is a sequence

cpuheater: I'm only refering to the update phase, inference is ok

vwxyzjn: So during the inference, it has
```
next_lstm_state[0].shape
Out[3]: torch.Size([1, 8, 128])
```
Isn't 8 the batch in this case?

cpuheater: yes, during inference you passs single sequence with a batch of 8, during update you concatenate two trajectories of length 128 to form single sequence of length 256 and batch of 1

cpuheater: for the update you need to switch batch and seq_length

vwxyzjn: I think I see wha you are saying, but shouldn’t it to have shape `(128, 2, 128)`? So basically the batch size is 2 and the sequence length is 128.

cpuheater: sure

vwxyzjn:
Ok, I have further looked into this a bit and have identified the following details:

1. Store the initial hidden state for the rollouts

openai/baselines' implementation does not store the hidden states for the rollouts, but rather it stores the hidden states at the start of the rollout.
That is the reason why all the hidden states' sum is zero in the first training iteration, as the screenshot shows below, 

image

So it would appear that openai/baselines' implementation is building the hidden states again given the initial hidden state


2. Apply mask inside of LSTM ([link](https://github.com/openai/baselines/blob/ea25b9e8b234e6ee1bca43083f8f3cf974143998/baselines/a2c/utils.py#L81))

While it builds the hidden states again given the initial hidden state, the episode termination mask is part of the LSTM module, which is **not the case** with PyTorch.

So I think what should have happened is we would have the following variables and shapes:

```
hidden, (128, 2, 512) (sequence length, batch size, input size) 
initial_lstm_state[0], (1, 2, 128) (num_LSTM_layers, batch size, hidden size) 
initial_lstm_state[1], (1, 2, 128) (num_LSTM_layers, batch size, hidden size) 

done, (128, 2, 512) (sequence length, batch size, input size) 
```

And we should do 
```
self.lstm(hidden, initial_lstm_state, mask)
```

Note that this API does not exist because PyTorch's LSTM module does not support this sort of mask.

Not sure how to proceed from here...

vwxyzjn:
OK a quick update on this, I think the original implementation is actually doing BPTT, so as to regenerate all the LSTM states from scratch during training. After very careful alignments and checking all the shapes, I have created another LSTM implementation.


See the latest reproduction graph here: https://wandb.ai/cleanrl/cleanRL/reports/LSTM-no-framestack--VmlldzoxMjEwNDI3. It seems that now we can closely match the performance of the original implementation.

A quick note to you @cpuheater is that we don't use the sequence length of the `nn.LSTM` because there is no way to implement the episode mask *within the sequence*. As such, we basically use a simple for loop to regenerate all the state again.

vwxyzjn:
image

Perfect reproduction. I will refactor the memory env stuff and think this PR will be ready to merge.

cleanrl/lstm/ppo_atari_lstm.py
```diff
@@ -0,0 +1,382 @@
...
+            nn.ReLU(),
+            layer_init(nn.Conv2d(32, 64, 4, stride=2)),
+            nn.ReLU(),
+            layer_init(nn.Conv2d(64, 64, 3, stride=1)),
+            nn.ReLU(),
+            nn.Flatten(),
+            layer_init(nn.Linear(64 * 7 * 7, 512)),
+            nn.ReLU(),
+        )
+        self.lstm = nn.LSTM(512, 128)
```
cpuheater, line 131: As you iterate over the sequence you could you LSTMCell instead.

vwxyzjn: I don’t quite follow this. Is LSTMCell a class?

cpuheater: Yes, the diffrence is that LSTM takes arbitrary number of steps whereas LSTMCell takes one step at the time.

cpuheater: And you passing a single step in the forward method, it's just a minor thing.

vwxyzjn:
We are able to match openai/baselines PPO + LSTM performance. See [runs here](https://wandb.ai/vwxyzjn/ppo-details/reports/Our-PPO-vs-openai-baselines-PPO--VmlldzoxMzE5MzYz) and the following screenshot
image


Also added an environment that requires memory to solve. The [run here](
https://wandb.ai/costa-huang/cleanRL/runs/73rohd17) shows PPO can learn to play this environment.

image

vwxyzjn:
Thanks, @dosssman, @yooceii, and @cpuheater  for reviewing this PR. Merging now.

PR #84, opened 2021-11-07
Rollback back PyTorch version for better compatibility

vwxyzjn:
`torch==1.10.0` is giving me some memory leak issues in gym-microrts... Hence rolling back to 1.7.1


vwxyzjn:
@dosssman Merging this now.

PR #90, opened 2021-11-15
Set the correct SPOT allocation strategy

vwxyzjn:
For some reason, the `BEST_FIT` strategy doesn't work and I cannot spin up any runs using the AWS spot computing environments. In this PR we change the allocation strategy to `SPOT_CAPACITY_OPTIMIZED`, which seems to be working.

image


dosssman:
I created a job definition matching the g4dn-xlarge-spot using the `BEST_FIT` and it still managed to run the job.
Not sure if it is relevant, but I ran the ppo_atari.py on Breakout as per the cloud documentation:
```bash
poetry run python -m cleanrl_utils.submit_exp \
    --docker-tag vwxyzjn/cleanrl:latest \
    --command "poetry run python cleanrl/ppo_atari.py --gym-id BreakoutNoFrameskip-v4 --track --capture-video" \
    --job-queue g4dn-xlarge \
    --num-seed 1 \
    --num-vcpu 1 \
    --num-gpu 1 \
    --num-memory 4000 \
    --num-hours 48.0 \
    --provider aws
```
and it crashed, around 150K steps.

`SPOT_CAPACITY_OPTIMIZED` also did work, but I terminated it early so I am not sure if it would also crash the same way as the one above.

vwxyzjn:
I see. Given this let’s put this PR on hold.

PR #91, opened 2021-11-18
Update paper citation entry

vwxyzjn:
This PR updates the BibTeX entry for CleanRL (https://arxiv.org/abs/2111.08819).


dosssman (approved):
:tada: :champagne:

PR #102, opened 2022-01-28
Refactor value based methods

vwxyzjn:
Continue on #79. This PR refactors the value based methods. Specifically, we

1. Adopted the replay buffer from Stable-baselines 3. SB3's replay buffer is well-implemented and its behavior is intuitive. That is, we can imagine what happens when we do `buffer.sample()` compared to something like `agent.learn()`, which is much more abstract. To this end, I think it's worth to adopt SB3's replay buffer. In the future, when SB3 introduce prioritized replay buffer (https://github.com/hill-a/stable-baselines/issues/751), it will also make it easier for us to adopt.
2. General format refactoring.

- [x] match DDPG performance
- [x] match TD3 performance
- [x] match C51 performance
- [x] match DQN performance

### DIFF
--- a/cleanrl/ddpg_continuous_action.py
+++ b/cleanrl/ddpg_continuous_action.py
@@ -6,7 +6,7 @@ from distutils.util import strtobool
 
 import gym
 import numpy as np
-# import pybullet_envs  # noqa
+import pybullet_envs  # noqa
 import torch
 import torch.nn as nn
 import torch.optim as optim
--- a/cleanrl/ddpg_continuous_action.py
+++ b/cleanrl/ddpg_continuous_action.py
@@ -1,144 +1,94 @@
-# Reference: https://arxiv.org/pdf/1509.02971.pdf
-# https://github.com/seungeunrho/minimalRL/blob/master/ddpg.py
-# https://github.com/openai/spinningup/blob/master/spinup/algos/pytorch/ddpg/ddpg.py
+import argparse
+import os
+import random
+import time
+from distutils.util import strtobool
 
+import gym
+import numpy as np
+# import pybullet_envs  # noqa
 import torch
 import torch.nn as nn
 import torch.optim as optim
 import torch.nn.functional as F
 from torch.utils.tensorboard import SummaryWriter
+from stable_baselines3.common.buffers import ReplayBuffer
 
-import argparse
-from distutils.util import strtobool
-import collections
-import numpy as np
-import gym
-import pybullet_envs
-from gym.wrappers import TimeLimit, Monitor
-from gym.spaces import Discrete, Box, MultiBinary, MultiDiscrete, Space
-import time
-import random
-import os
+def parse_args():
+    # fmt: off
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--exp-name", type=str, default=os.path.basename(__file__).rstrip(".py"),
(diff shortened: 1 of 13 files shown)


vwxyzjn:
Some benchmarked experiments: https://wandb.ai/gym-microrts/gym-microrts/runs/2m3m2b1b?workspace=user-costa-huang

vwxyzjn:
Found a regression with C51  probably due to different learning rate and target-network-frequency

image

vwxyzjn:
Found a regression with DQN, probably due to different learning rate

image

cleanrl/ddpg_continuous_action.py
```diff
@@ -148,95 +98,132 @@ def forward(self, x, a, device):
...
+
+    # env setup
+    envs = gym.vector.SyncVectorEnv(
+        [make_env(args.gym_id, 0, 0, args.capture_video, run_name)]
+    )
+    assert isinstance(envs.single_action_space, gym.spaces.Box), "only continuous action space is supported"
+
+    max_action = float(envs.single_action_space.high[0])
+    actor = Actor(envs).to(device)
+    qf1 = QNetwork(envs).to(device)
```
dosssman, line 148: Probably a bit nit-picky, but everywhere there is a need to create a copy, such as `qf1_target` for example, we could simply use `qf1_target = qf1.clone()`, making it slightly cleaner.

Additionally, disabling gradients with `qf1.clone().requires_grad_(False)` would make it slightly more memory efficient I think, but the improvement is probably marginal.

vwxyzjn: That's a good point. Although I would probably keep it as-is because it makes the device `qf1` belongs to more explicit. Also, we were using `torch.no_grad()` so the gradient wouldn't accumulate anyway.

vwxyzjn:
The regression report checks through 

image

@dosssman all good on your end?

cleanrl/ddpg_continuous_action.py
dosssman: Actually, it wasn't `.clone()` but `copy.deepcopy()`. Got bamboozled by my memory somehow.
Let's keep it as it is.

dosssman:
I have been looking at the continous action space methods which I am more familiar with.
I think there was some superfluous code in the TD3 scripts, which is addressed in #106 .
Also added refactored SAC with some additional tweaks that could be incorporated, while we are at it.

Will further check the DQN / C51 like discrete action space further in the week.

In any case, great work as always.

vwxyzjn:
Thanks @dosssman for the detailed check on TD3, after this fix, TD3's performance (green line on the right) is even better

image

dosssman:
Great. Thanks for running the experiments to check those.
On my side I have only tested SAC refactored (orange line) and the previous version (blue line), albeit with only one seed.
image

By the way, is the wandb report which the screenshot above come from open ? I could not find it in the cleanrl project so far.

vwxyzjn:
It's here: https://wandb.ai/costa-huang/cleanRL/reports/Regression-Report--VmlldzoxNDI1MTE4

vwxyzjn:
Merging as is so that I could introduce formatting piplines after discussing with @dosssman (great work btw). And if there are remaining issues we can open new PRs.

PR #109, opened 2022-02-05
Add a profiler example

vwxyzjn:
This PR adds a profiler example with PPO and Atari leveraging [kineto](https://github.com/pytorch/kineto/tree/main/tb_plugin).

## How to run

```bash
https://github.com/vwxyzjn/cleanrl.git
git checkout profile
poetry install
poetry run python cleanrl/ppo_atari_profile.py
tensorboard --logdir runs
``` 

Then go to the `PYTORCH_PROFILER` tab in the tensorboard UI as follows:

https://user-images.githubusercontent.com/5555347/152648445-4257ab3b-5b8e-409c-8825-21fe5751aedd.mp4

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ mjkey.txt
 __pycache__/
 *.py[cod]
 *$py.class
+*.pt.trace.json
 
 # C extensions
 *.so
--- a/cleanrl/ppo_atari_profile.py
+++ b/cleanrl/ppo_atari_profile.py
@@ -210,7 +210,6 @@ if __name__ == "__main__":
     ) as prof:
 
         for update in range(1, num_updates + 1):
-
             # Annealing the rate if instructed to do so.
             if args.anneal_lr:
                 frac = 1.0 - (update - 1.0) / num_updates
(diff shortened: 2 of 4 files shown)


vwxyzjn:
Some interesting highlights:

1. In the `Execution Summary`, the `CPU Exec` takes 55.7% of the time, which likely suggests the environment's steps are still quite expensive.
1. In the `Execution Summary`, the `Runtime` takes 38.4% of the time. According to the [docs](https://github.com/pytorch/kineto/tree/main/tb_plugin), `Runtime` is 
   >  Runtime: CUDA runtime execution time on host side; Such as cudaLaunchKernel, cudaMemcpyAsync, cudaStreamSynchronize, ...
   * I think this means 1) moving observations from CPU to GPU to sample actions and 2)  moving sampled actions from GPU to CPU to step environments are expensive
1. In the `Execution Summary`, the `Kernel` (the actual GPU accelerated operations) only takes 4.52% of the time.

vwxyzjn:
Annotated things more clearly, which results in the following profile

image

vwxyzjn:
Tracked run here:

https://wandb.ai/costa-huang/cleanRL/runs/2nd904gd/tensorboard?workspace=user-costa-huang

Turns out the tensorboard integration is working when I manually upload the profiler's trace 

image

Issue #115, opened 2022-02-21
Roadmap for CleanRL


vwxyzjn:
As CleanRL gets more mature, it's time to re-think the future. With CleanRL 1.0, we'd hope to further improve documentation and design better contribution guidelines. This issue tracks a few desired items for CleanRL 1.0.

## 1.0
- [x] #116
- [x] Table of implemented algorithms and links that refer to the paper. (see https://docs.cleanrl.dev/rl-algorithms/overview/)
- [x] #121
- [x] #123
- [x] #117
- [x] #171 
- [x] #172 
- [x] #168 
- [x] #148
- [x] #158 

## 1.1+
- [ ] #228
- [ ] #265
- [ ] #184
- [ ] #183
- [x] #127
- [x] #196
- [x] #206
- [x] #218
- [x] #263

PR #118, opened 2022-02-21
Refactor `parse_args()`

vwxyzjn:
This PR refactors a few arguments in `parse_args()`:

1. This PR moves the `gym_id`, `learning_rate`, `total_timesteps` move down to algorithm-specific arguments. Currently when doing a file diff between `c51_atari.py` and `ppo_continuous_action.py`, we have the following screenshot:
image
However, it's more desirable to all of the differences in one place, as shown below:
image

2. This PR also renames the `gym-id` to `env-id`. The reason is that as we adopt different environments, some of them don't necessarily have a `gym-id` but they have an environment name. For example, procgen has the `env-id` but not `gym-id`.

Curious about your thoughts @yooceii @dosssman on whether this is necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ poetry install
 # `python run cleanrl/ppo.py`
 poetry run python cleanrl/ppo.py \
     --seed 1 \
-    --gym-id CartPole-v0 \
+    --env-id CartPole-v0 \
     --total-timesteps 50000
 
 # open another temrminal and enter `cd cleanrl/cleanrl`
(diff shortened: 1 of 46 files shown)


dosssman (approved):
I concur on the `gym-id` to `env-id` change, as the former is more general indeed.

For the position of the arguments in the argparser list, since some algorithms are kind of made for speicfic tasks (SAC -> Bulet / Mujoco), it does make sense to put it in the algorithms specific arguments. `learning-rate` too, because in SAC for example, there are actually 2 learning, so it would make the parameterization cleaner.

While total-timesteps is also general, I guess different algorithms training for different amount of steps could be possible (for example Atari's required 10M + vs Bullet's 1M) can be used as justification for this change.

In any case, it is all good on my side.

vwxyzjn:
The matplotlib utilities might break but that's ok since we are re-building Open RL Benchmark anyway #115.

PR #120, opened 2022-02-22
Add PPO documetnation

vwxyzjn:


vwxyzjn:
Merging as is but happy to revisit.

PR #125, opened 2022-02-28
Replace `episode_reward` with `episodic_return`

vwxyzjn:
This PR replaces some remaining references to `episode_reward` with `episodic_return`. `episodic_return` is a more accurate term: it describes the sum of rewards obtained within an episode. In comparison, the term `episode_reward` is ambiguous and therefore should be replaced.


vwxyzjn:
Merging now.

PR #126, opened 2022-02-28
Add SPS and q-values metrics for value-based methods

vwxyzjn:
This PR adds SPS and q-values metrics for value methods. 

This PR also use `F.mse_loss()` instead of `loss_fn = nn.MSELoss()`

image
image

@dosssman do you think we should also use `using qf1() instead, which calls forward() by default`?

### DIFF
--- a/cleanrl/apex_dqn_atari.py
+++ b/cleanrl/apex_dqn_atari.py
@@ -776,7 +776,7 @@ def act(args, experiment_name, i, q_network, target_network, lock, rollouts_queu
         # ALGO LOGIC: put action logic here
         epsilon = linear_schedule(args.start_e, args.end_e, args.exploration_fraction * args.total_timesteps, global_step)
         obs = np.array(obs)
-        logits = q_network.forward(obs.reshape((1,) + obs.shape), device)
+        logits = q_network(obs.reshape((1,) + obs.shape), device)
         if args.capture_video:
             if i == 0:
                 env.set_q_values(logits.tolist())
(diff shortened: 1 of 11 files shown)


dosssman:
@vwxyzjn Regarding the `qf1() instead of qf1.forward()` I have a subjective preference because it just looks "cleaner" to me, and that is what I use in my code bases.

One potential downside that comes to mind however, is that a newcomer that does not know this is the default behavior of pytorch might get confused at first.
Using `qf1.forward()`, it could be easier for the reader to realize that this operation is used the execute the logic coded in the `def forward(self, ...)` of the Qnetwork class.

Although in SB3, they do seem to use the `qf1()` way I think:
https://github.com/DLR-RM/stable-baselines3/blob/cdaa9ab418aec18f41c7e8e12e0ad28f238553eb/stable_baselines3/common/torch_layers.py#L233

In any case, I floated this change in sac.py last time, but still a little bit on the fence on how appropriate it would be for `cleanrl`.

vwxyzjn:
Fair enough, let's go ahead with this change `qf1() instead of qf1.forward()`.

vwxyzjn:
@dosssman  it's ready for review.

dosssman (approved):
Sorry for the late answer. The review request escaped me somehow.

First, I would like to confirm the meaning of SPS.
(Sample || Step per seconds ?)

I have compiled a few minor tweaks that build on top of this PR in #130 .
Hopefully I did not misunderstand the intent of this PR.

### Some additional comments are:
- This might just be my OCD, but when printing the SPS for the terminal and tensorboard / wandb logging like here for example:
https://github.com/vwxyzjn/cleanrl/blob/bb8dd13f3c7fd48a3da09cfff12059b6416dcb2c/cleanrl/dqn_atari.py#L213-L214

how about doing something like this ?
```python
SPS = int(global_step / (time.time() - start_time))
print("SPS:", int(global_step / (time.time() - start_time)))
writer.add_scalar("charts/SPS", int(global_step / (time.time() - start_time)), global_step)
```
Furthermore, we could also fuse the SPS and episode return logging here to only print once? :
https://github.com/vwxyzjn/cleanrl/blob/bb8dd13f3c7fd48a3da09cfff12059b6416dcb2c/cleanrl/dqn_atari.py#L184-L189

- The update to gym=0.23.0 breaks a few scripts, namely the `rnd_ppo.py` and the 'offline/*` scripts due to the `Monitor` gym wrapper being removed form that version. This is the reason #130 tests break.

### Regarding `cleanrl/offline`
- SPS logging is missing
- Monitor cannot be imported anymore due to gym=0.23.0 update
- Tests for those two scripts are missing
- Unlike `dqn_atari`, the wrappers are not imported from SB3
- The offline-env-id that is required to load the dataset does not seem to work anymore. Is there any dependency missing, such as `d4rl` or `d4rl_atari` for example ?

Since the `offline` scripts are not really related to this PR, I did not go too much in detail.
We could merge this PR while opening an issue for the offline scripts, if not already done.

In any case, great job as always.

vwxyzjn:
Merging now.

PR #114, opened 2022-02-19
Update to `Gym==0.22.0`

vwxyzjn:
After 114 Minutes, the lock operation finished...

image

PR #129, opened 2022-03-08
Update gym to 0.23.0

vwxyzjn:
Follow up from #114.


vwxyzjn:
Tests pass. Merging now.

Issue #127, opened 2022-02-28
Add `rnd_ppo.py` documentation and refactor


vwxyzjn:
`rnd_ppo.py` is a bit dated, and I recommend refactoring it to match other PPO style, which would include:

- [x] change the name from `rnd_ppo.py` to `ppo_rnd.py`
- [x] use `from gym.wrappers.normalize import RunningMeanStd` instead of the implementing ourselves (note the implementation might be a bit different).
- [x] create a `make_env` function like https://github.com/vwxyzjn/cleanrl/blob/0b3f8eae7d07b90a0ee129ffe290bd82e5b57a14/cleanrl/ppo_atari.py#L88-L103
- [x] remove the visualization (i.e., `ProbsVisualizationWrapper`)
- [x] use ` def get_value` and `def get_action_and_value` for the `Agent` class
- [x] remove https://github.com/vwxyzjn/cleanrl/blob/0b3f8eae7d07b90a0ee129ffe290bd82e5b57a14/cleanrl/rnd_ppo.py#L706-L708
- [x] maybe log the average `curiosity_reward` instead? https://github.com/vwxyzjn/cleanrl/blob/0b3f8eae7d07b90a0ee129ffe290bd82e5b57a14/cleanrl/rnd_ppo.py#L848
- [x] name `total_reward_per_env` to `curiosity_return` https://github.com/vwxyzjn/cleanrl/blob/0b3f8eae7d07b90a0ee129ffe290bd82e5b57a14/cleanrl/rnd_ppo.py#L854
- [x] Add SPS (steps per second) metric. 

Overall I suggest selecting `ppo_atari.py` and `rnd_ppo.py` and use `Compare Selected` on VSCode to see the file difference and minimize the file difference:

![image](https://user-images.githubusercontent.com/5555347/156015137-e0b6b85c-0f61-4e03-a1e5-c7ff7e044cb4.png)

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] New algorithm
- [x] Documentation

## Checklist:

- [x] I've read the [CONTRIBUTION](https://github.com/vwxyzjn/cleanrl/blob/master/CONTRIBUTING.md) guide (**required**).
- [x] I have ensured `pre-commit run --all-files` passes (**required**).
- [x] I have updated the documentation accordingly.
- [ ] I have updated the tests accordingly (if applicable).

If you are adding new algorithms or your change could result in performance difference, you may need to (re-)run tracked experiments. 
[...]

PR #130, opened 2022-03-09
Tweaks for PR 'add-sps-qvalues'

dosssman:
###  apex_dqn_atari..py
- Added SPS logging to match other files in #126
- Added `apex_dqn_atari.py` to `test_atari.py`

### SAC Continuous
- Removed the comment pertaining to (1) using F.mse_loss() and (2) using qf1() instead of qf1.forward()

### rnd_ppo:
- removed `.forward()` calls to match the other files in #126 
- Added to `test_atari.py`

### DIFF
--- a/cleanrl/rnd_ppo.py
+++ b/cleanrl/rnd_ppo.py
@@ -692,14 +692,14 @@ class Agent(nn.Module):
         return self.network(x)
 
     def get_action(self, x, action=None):
-        logits = self.actor(self(x))
+        logits = self.actor(self.forward(x))
         probs = Categorical(logits=logits)
         if action is None:
             action = probs.sample()
         return action, probs.log_prob(action), probs.entropy()
 
     def get_value(self, x):
-        features = self(x)
+        features = self.forward(x)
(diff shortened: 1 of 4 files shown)


dosssman:
Regarding the check that fails for `rnd_ppo.py`, it is likely due to the update to `gym=0.22.0`, which removes the 'Monitor' wrapper class the scripts depends on.

cleanrl/rnd_ppo.py
```diff
@@ -692,14 +692,14 @@ def forward(self, x):
...
     def get_action(self, x, action=None):
-        logits = self.actor(self.forward(x))
+        logits = self.actor(self(x))
         probs = Categorical(logits=logits)
         if action is None:
             action = probs.sample()
         return action, probs.log_prob(action), probs.entropy()
 
     def get_value(self, x):
-        features = self.forward(x)
```
vwxyzjn, line 702: RND needs to be completely refactored: see #127. I would revert changes here because we would no longer use a `forward` function in `rnd_ppo.py and instead use a `get_action_and_value` Function.

vwxyzjn (changes requested):

vwxyzjn:
Could you also fix the CI please?

PR #134, opened 2022-03-09
Make seed work again in value methods

### LINKED ISSUE #133: `dqn.py` does not respect seed
## Problem Description

`python dqn.py --seed 1` could yield different results. See the following demo, where the first run yields `global_step=24830, episodic_return=245.0` and the second run yields `global_step=24975, episodic_return=178.0`.

[asciicast](https://asciinema.org/a/475396)

vwxyzjn:
Closes #133. Here is a demo for the fix:

[asciicast](https://asciinema.org/a/475401)


vwxyzjn:
Merging now.

PR #137, opened 2022-03-13
Add `ddpg_continuous_action.py` docs

vwxyzjn:
This RP adds docs for `ddpg_continuous_action.py`. 

## Checklist for `ddpg_continuous_action.py`:

- [x] I've read the [CONTRIBUTION](https://github.com/vwxyzjn/cleanrl/blob/master/CONTRIBUTING.md) guide (**required**).
- [x] I have ensured `pre-commit run --all-files` passes (**required**).
- [x] I have contacted @vwxyzjn to obtain access to the [openrlbenchmark W&B team](https://wandb.ai/openrlbenchmark) (**required**).
- [x] I have tracked applicable experiments in [openrlbenchmark/cleanrl](https://wandb.ai/openrlbenchmark/cleanrl) with `--capture-video` flag toggled on (**required**).
- [x] I have updated the documentation and previewed the changes via `mkdocs serve`.
    - [x] I have explained note-worthy implementation details.
    - [x] I have explained the logged metrics.
    - [x] I have added links to the original paper and related papers (if applicable).
    - [x] I have added links to the PR related to the algorithm.
    - [x] I have created a table comparing my results against those from reputable sources (i.e., the original paper or other reference implementation).
    - [x] I have added the learning curves (in PNG format with `width=500` and `height=300`).
    - [x] I have added links to the tracked experiments.
- [x] I have updated the tests accordingly (if applicable).

PR #157, opened 2022-04-09
Add docs for `dqn.py`

vwxyzjn:
## Description
This PR adds documentation for `dqn.py` and couple documentation fixes for `dqn_atari.py`

## Types of changes

- [x] Documentation

## Checklist:

    - [x] I have added links to the original paper and related papers (if applicable).
    - [x] I have added links to the PR related to the algorithm.


vwxyzjn:
This PR is ready for review.

vwxyzjn:
Additional note: I realized it would be pretty hard to run benchmark experiments based on some master's version. From now on, we should do something like

```
gh pr checkout 157
poetry install
```

in the reproducibility instructions.

cleanrl/dqn.py
```diff
@@ -49,17 +49,17 @@ def parse_args():
...
         help="the starting epsilon for exploration")
     parser.add_argument("--end-e", type=float, default=0.05,
         help="the ending epsilon for exploration")
-    parser.add_argument("--exploration-fraction", type=float, default=0.8,
+    parser.add_argument("--exploration-fraction", type=float, default=0.5,
         help="the fraction of `total-timesteps` it takes from start-e to go end-e")
     parser.add_argument("--learning-starts", type=int, default=10000,
         help="timestep to start learning")
-    parser.add_argument("--train-frequency", type=int, default=1,
+    parser.add_argument("--train-frequency", type=int, default=10,
```
dosssman, line 62: Shouldn't DQN update after each environemnt step ?

dosssman (commented):
Looking good over here too.
I noticed there is no dqn_atari.py benchmarks in the docs though. Will it be alright ?

cleanrl/dqn.py
vwxyzjn: See the DQN paper

Screen Shot 2022-04-19 at 10 38 45 PM
 
And in C51 paper: 

"While our training regime closely follows that of DQN (Mnih et al., 2015),..."

vwxyzjn:
> Looking good over here too. I noticed there is no dqn_atari.py benchmarks in the docs though. Will it be alright ?

The `dqn_atari.py` docs are already added from #124.

vwxyzjn:
Let me know if this addresses your concern :)

cleanrl/dqn.py
dosssman: I see. Why 10 in this case ?

vwxyzjn: Just a tuned one tbh

dosssman: Since it works never the less, looks good to me.

dosssman (approved):

cleanrl/dqn.py
vwxyzjn: Another motivation for this is to collect more data before learning, similar to how it works in ppo. But yeah it’s just a hyper parameter and seems to work well.

PR #162, opened 2022-04-16
Prototype multi-gpu support with PPO

vwxyzjn:
## Description

This PR contains some prototypes that bring multi-GPU support to PPO. There are many ways to do it so this PR tries to compare different approaches.

I don't really have multi-GPUs to test this out, so I launch two processes accessing the same GPU, mainly to test out if they would result in the same performance. In theory, multi-GPU support should not harm the performance. However, I plan to real multi-GPU performance on the sample efficiency is not affected.

### Option 1: `ppo_atari_multigpu.py`

My first try is `ppo_atari_multigpu.py`, which uses pytorch's low-level distributed API as shown in [this link](https://pytorch.org/tutorials/intermediate/dist_tuto.html#distributed-training) or this example:

```python
""" Gradient averaging. """
def average_gradients(model):
    size = float(dist.get_world_size())
    for param in model.parameters():
        dist.all_reduce(param.grad.data, op=dist.ReduceOp.SUM)
        param.grad.data /= size
```

### Option 2 `ppo_atari_multigpu_batch_reduce.py`

In this file, I adopted https://github.com/entity-neural-network/incubator/pull/220 by batch reducing the gradient.

### Option 3 `ppo_atari_ddp.py`

In this file, I adopted the high-level API here https://pytorch.org/docs/stable/notes/ddp.html.

### Option 4 `ppo_atari_elastic.py`

This file adopts https://pytorch.org/docs/stable/elastic/run.html. We can run the training script via `torchrun --standalone --nnodes=1 --nproc_per_node=2 ppo_atari_elastic.py`

image

The sample efficiency seems to suffer, as shown above, however the wall-time performance is pretty good.

My suspicion for the sample efficiency regression is that policy gradient averaging is more tricky compared to the value gradient averaging: see https://github.com/vwxyzjn/cleanrl/pull/162#issuecomment-1100796257

Options 3 and 4 are pretty impressive: they reduce the wall-time by half while using only a single GPU: maybe by using multi-GPU the speed up can be even more?

## Types of changes

- [x] New feature
- [x] New algorithm
- [ ] Documentation

## Checklist:

- [ ] I have contacted @vwxyzjn to obtain access to the [openrlbenchmark W&B team](https://wandb.ai/openrlbenchmark) (**required**).


vwxyzjn:
Here is a script for understanding how gradient accumulation works and data parallelism works:

```python
import torch.distributed as dist
import torch.multiprocessing as mp
import os
import torch

def init_process(rank, size, fn, backend="gloo"):
    """Initialize the distributed environment."""
    os.environ["MASTER_ADDR"] = "127.0.0.1"
    os.environ["MASTER_PORT"] = "29500"
    dist.init_process_group(backend, rank=rank, world_size=size)
    fn(rank, size)

def train(rank: int, size: int):
    prediction = torch.tensor(
        [[1.,2.,3.,4.], [4.,7.,5.,8.]],
    requires_grad=True)
    label = torch.tensor([[0.,0.,0.,0.], [0.,0.,0.,0.]])
    loss = (prediction[rank] - label[rank]) ** 2
    loss.mean().backward()
    dist.all_reduce(prediction.grad.data, op=dist.ReduceOp.SUM)
    prediction.grad.data /= size
    print("gradient with data parallelism (multi-gpu) \n", prediction.grad)
    

if __name__ == "__main__":
    prediction = torch.tensor(
        [[1.,2.,3.,4.], [4.,7.,5.,8.]],
    requires_grad=True)
    label = torch.tensor([[0.,0.,0.,0.], [0.,0.,0.,0.]])

    loss = (prediction - label) ** 2
    loss.mean().backward()
    print("gradient with the whole batch\n", prediction.grad)

    prediction = torch.tensor(
        [[1.,2.,3.,4.], [4.,7.,5.,8.]],
    requires_grad=True)
    label = torch.tensor([[0.,0.,0.,0.], [0.,0.,0.,0.]])
    # do backward pass in two minibatches
    loss = (prediction[0] - label[0]) ** 2
    loss.mean().backward()
    loss = (prediction[1] - label[1]) ** 2
    loss.mean().backward()
    # divide the gradient by the size
    print("gradient accumulation \n", prediction.grad / 2)
    
    size = 2
    processes = []
    mp.set_start_method("spawn")
    for rank in range(size):
        p = mp.Process(target=init_process, args=(rank, size, train))
        p.start()
        processes.append(p)

    for p in processes:
        p.join()
```

the script yields the following output:

```
gradient with the whole batch
 tensor([[0.2500, 0.5000, 0.7500, 1.0000],
        [1.0000, 1.7500, 1.2500, 2.0000]])
gradient accumulation 
 tensor([[0.2500, 0.5000, 0.7500, 1.0000],
        [1.0000, 1.7500, 1.2500, 2.0000]])
gradient with data parallelism (multi-gpu) 
 tensor([[0.2500, 0.5000, 0.7500, 1.0000],
        [1.0000, 1.7500, 1.2500, 2.0000]])
gradient with data parallelism (multi-gpu) 
 tensor([[0.2500, 0.5000, 0.7500, 1.0000],
        [1.0000, 1.7500, 1.2500, 2.0000]])
```

vwxyzjn:
`ppo_atari.py` does a single forward and backward pass on a minibatch of size 256, whereas `ppo_atari_multigpu.py` does two forward and backward passes in two separate processes on two minibatches of size 128, call `dist.all_reduce(param.grad.data, op=dist.ReduceOp.SUM)` to sum the gradient then divide the gradient by the number of processes 2.

The following video in principal demonstrates what `ppo_atari_multigpu.py` does. At the end of the video, we see that the gradient is the same for the critic layer when 1) doing a single forward pass and 2) doing two separate forward passes.

https://user-images.githubusercontent.com/5555347/163699097-5128b64f-3d6f-45d9-9a8d-5da7263c1574.mov

However, we could not get the same gradient when with the policy layer... Going to investigate this further.

https://user-images.githubusercontent.com/5555347/163699209-b380cb10-049e-4283-86d3-b6d8eadf4d31.mov

vwxyzjn:
Here is an even simpler demo. The issue is `Categorical.log_prob` does not work with gradient accumulation.

```
import torch
import torch.nn as nn
import torch.optim as optim
from torch.distributions.categorical import Categorical

class Agent(nn.Module):
    def __init__(self, action_n):
        super().__init__()
        self.network = nn.Sequential(
            nn.Conv2d(4, 32, 8, stride=4),
            nn.ReLU(),
            nn.Conv2d(32, 64, 4, stride=2),
            nn.ReLU(),
            nn.Conv2d(64, 64, 3, stride=1),
            nn.ReLU(),
            nn.Flatten(),
            nn.Linear(64 * 7 * 7, 512),
            nn.ReLU(),
        )
        self.actor = nn.Linear(512, action_n)
        self.critic = nn.Linear(512, 1)

    def get_value(self, x):
        return self.critic(self.network(x / 255.0))

    def get_action_and_value(self, x, action=None):
        hidden = self.network(x / 255.0)
        logits = self.actor(hidden)
        if action is None:
            action = probs.sample()
        return action, probs.log_prob(action), probs.entropy(), self.critic(hidden)

# setup
agent = Agent(4)
optimizer = optim.Adam(agent.parameters())
next_obs = torch.rand(8, 4, 84, 84)
action, newlogprob, entropy, newvalue = agent.get_action_and_value(next_obs)

optimizer.zero_grad()
_, newlogprob, _, newvalue = agent.get_action_and_value(next_obs, action)
newvalue.mean().backward()
print(f"`agent.critic.weight.grad.sum() = {agent.critic.weight.grad.sum()}` after fitting value loss using data[0:8]")

optimizer.zero_grad()
_, newlogprob, _, newvalue = agent.get_action_and_value(next_obs[0:4], action[0:4])
newvalue.mean().backward()
_, newlogprob, _, newvalue = agent.get_action_and_value(next_obs[4:8], action[4:8])
newvalue.mean().backward()
print(f"`agent.critic.weight.grad.sum() / 2 = {agent.critic.weight.grad.sum() / 2}` after fitting value loss using data[0:4] and data[4:8] respectively")

optimizer.zero_grad()
_, newlogprob, _, newvalue = agent.get_action_and_value(next_obs, action)
newlogprob.mean().backward()
print(f"`agent.actor.weight.grad.sum() = {agent.actor.weight.grad.sum()}` after fitting value loss using data[0:8]")

optimizer.zero_grad()
_, newlogprob, _, newvalue = agent.get_action_and_value(next_obs[0:4], action[0:4])
newlogprob.mean().backward()
_, newlogprob, _, newvalue = agent.get_action_and_value(next_obs[4:8], action[4:8])
newlogprob.mean().backward()
print(f"`agent.actor.weight.grad.sum() / 2 = {agent.actor.weight.grad.sum() / 2}` after fitting value loss using data[0:4] and data[4:8] respectively")
```

vwxyzjn:
## Attempt 2: much more successful

image

Fixed a couple of issues:
* In `ppo_atari_multigpu.py` and `ppo_atari_multigpu_batch_reduce.py`, I needed to set different environment random seeds used in separate processes. Previously I didn't and thus the script is essentially only learning from half of the experience.
* In `ppo_atari_ddp.py` and `ppo_atari_elastic.py`, for some reason `agent.module.get_action_value` does not work with multiprocesses. I needed to rename `get_action_value` to `forward` to make things work.

cleanrl/ppo_atari_dpp.py
```diff
@@ -0,0 +1,372 @@
...
+            action = probs.sample()
+        return action, probs.log_prob(action), probs.entropy(), self.critic(hidden)
+
+
+def train(rank: int, size: int):
+    args = parse_args()
+    args.num_envs = int(args.num_envs / size)
+    args.batch_size = int(args.num_envs * args.num_steps)
+    args.minibatch_size = int(args.batch_size // args.num_minibatches)
+    dist.init_process_group("gloo", rank=rank, world_size=size)
```
araffin, line 157: According to PyTorch doc, you should be using NCCL for multi-gpu training, Gloo is recommended for multi-cpu training: https://pytorch.org/docs/stable/distributed.html

vwxyzjn: I don’t actually have a multi GPU stuff to try it out and NCCL would break :) but this is something I should try find a solution for.

vwxyzjn:
48C88F0D-C3A9-4EC7-853D-2E216DB9DBEB


https://wandb.ai/costa-huang/cleanRL/reports/Data-Parallelism-Experiment--VmlldzoxODI1OTY0

Ok did more testing and it looks like `ppo_atari_multigpu_batch_reduce.py` has the highest performance.

I do need to find a multi-GPU machine to test this out though

vwxyzjn:
Closing in favor of #178

PR #199, opened 2022-06-10
Add a note on PPG's performance

vwxyzjn:
## Description
This PR adds some documentation explaining why our PPG's performance is different from the original PPG's results.

## Types of changes
- [x] Documentation


Dipamc77:
Looks good to me.

PR #200, opened 2022-06-10
Clarify CleanRL is a non-modular library

### LINKED ISSUE #197: CleanRL can't be used by importing?
Hi @vwxyzjn !

Jordan recommended that I use CleanRL when working with Petting Zoo, and it took me a bit to figure out something. Let me know whether I understand correctly. CleanRL isn't meant to be imported and used directly, like SB3. It's supposed to have reference implementations of algorithms which developers could copy off to their own program and tweak as necessary to serve their needs. 

Is this correct? If so:

1. It should be clearly stated in the readme, even in the title. Using the phrase "reference implementations" is helpful, and there needs to be at least a sentence that explains that CleanRL isn't meant to be imported, it's meant to be copied from. 
2. It would be cool if the implementations in CleanRL would have been useful in importable forms in some way. It could be a separate package. Many people want to just `import PPO` and have it work.

vwxyzjn:
## Description
Closes #197.

## Types of changes
- [x] Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ CleanRL is a Deep Reinforcement Learning library that provides high-quality sing
 
 * 📜 Single-file implementation
    * *Every detail about an algorithm variant is put into a single standalone file.* 
-   * For example, our `ppo_atari.py` only has 340 lines of code but contains all implementation details on how PPO works with Atari games, so it is a great reference implementation to read for folks who do not wish to read an entire modular library.
+   * For example, our `ppo_atari.py` only has 340 lines of code but contains all implementation details on how PPO works with Atari games, **so it is a great reference implementation to read for folks who do not wish to read an entire modular library**.
 * 📊 Benchmarked Implementation (7+ algorithms and 34+ games at https://benchmark.cleanrl.dev)
 * 📈 Tensorboard Logging
 * 🪛 Local Reproducibility via Seeding
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ You can read more about CleanRL in our [technical paper](https://arxiv.org/abs/2
 
 Good luck have fun :rocket:
 
-⚠️ **NOTE**: CleanRL is *not* a modular library and therefore it is not meant to be imported. At the cost of duplicate code, we make all implementation details of a DRL algorithm variant easy to understand, so CleanRL comes with its own pros and cons. You should consider using CleanRL if you want to 1) understand all implementation details of an algorithm's varaint or 2) prototype advanced features that other modular DRL libraries do not support (CleanRL has minimal lines of code so it gives you great debugging experience and you don't have do a lot of subclassing like somtimes in modular DRL libraries).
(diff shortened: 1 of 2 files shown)


docs/index.md
```diff
@@ -13,19 +13,21 @@
...
+* 🎮 Videos of Gameplay Capturing
+* 🧫 Experiment Management with [Weights and Biases](https://wandb.ai/site)
+* 💸 Cloud Integration with docker and AWS 
 
 You can read more about CleanRL in our [technical paper](https://arxiv.org/abs/2111.08819) and [documentation](https://docs.cleanrl.dev/).
 
-Good luck have fun 🚀
+Good luck have fun :rocket:
+
+⚠️ **NOTE**: CleanRL is *not* a modular library and therefore it is not meant to be imported. At the cost of duplicate code, we make all implementation details of a DRL algorithm variant easy to understand, so CleanRL comes with its own pros and cons. You should consider using CleanRL if you want to 1) understand all implementation details of an algorithm's varaint or 2) do quick prototypes.
```
cool-RR, line 30: varaint -> variant

I'm not sure that "do quick prototypes" makes sense here. Running `from clearrl import PPO` would be quick. Reading the algorithm and copy-pasting it into my code is slow.

cool-RR (approved):
Nice. Personally I'd put the phrase "reference implementation" into the title of the readme and the  "about" of the repo (where it says "High-quality single file implementation".) But it's your decision.

cool-RR:
Lost interest?

docs/index.md
vwxyzjn: I think "doing prototypes" is not really a well-defined notion as there are many types of prototypes. Being able to do prototypes quickly largely depends on the use case. 

While things like `from stable_baselines3 import PPO` is quick but if you want to prototype advanced features that SB3 does not support, it could be more difficult as discussed in #197 with the invalid action masking example.

Maybe I can clarify as "do prototypes that can't be achieved by just combining components in modular DRL libraries"? I am really unsure what the phrasing would be.

vwxyzjn:
Hey sorry for the delay. This week is my end-of-internship and a move-out week. 

Yes, let me think of a place to put the reference implementation. Please let me know what you think of my reply above regarding the prototypes comment.

README.md
```diff
@@ -16,7 +16,8 @@ CleanRL is a Deep Reinforcement Learning library that provides high-quality sing
 
 
 * 📜 Single-file implementation
-   * *Every detail about an algorithm is put into the algorithm's own file.* It is therefore easier to fully understand an algorithm and do research with.
+   * *Every detail about an algorithm variant is put into a single standalone file.* 
+   * For example, our `ppo_atari.py` only has 340 lines of code but contains all implementation details on how PPO works with Atari games, so it is a great reference implementation to read for folks who do not wish to read an entire modular library.
```
vwxyzjn, line 20: “reference implementation” is already mentioned here. The reason I am hesitant to put it in the title is because our paper already has a title that does not have the “reference” word in it, so I’d avoid changing it.

PR #205, opened 2022-06-20
Show correct exception cause

cool-RR:
For more context, see https://blog.ram.rachum.com/post/621791438475296768/improving-python-exception-chaining-with


vwxyzjn:
Thank you! This makes sense.

Issue #210, opened 2022-06-20
Adding Average Reward PPO proposal


Howuhh:
Although it is now common to solve most problems using discounted reward, this does not always correspond to the real problem (not episodic, long-horizon), where it is important to use algorithms that optimize the average reward. 

There are only two adaptations of modern algorithms for average-reward setting: [A-TRPO](https://proceedings.mlr.press/v139/zhang21q/zhang21q.pdf), [A-PPO](https://arxiv.org/pdf/2106.03442.pdf).  A-PPO is almost the same as regular PPO and much easier to implement than A-TRPO. Besides, it also solves some problems [[1](https://arxiv.org/pdf/1906.07073.pdf), [2](https://arxiv.org/pdf/1910.02140.pdf)] of regular PPO (e.g. sampling from undiscounted state distribution).

In my free time, I tried to [reproduce](https://github.com/Howuhh/average_reward_ppo) the results of the paper and it kinda worked. Nevertheless, I think it is important to compare it with the other algorithms, esp PPO. This will be easier to do if they share the same code base and common hacks. So it seems to me that adding it to cleanrl will help with this. In addition, I think average reward setting is very underrated and this will help popularize it (if A-PPO really works).

Howuhh:
Yes, but it will take some time, especially documentation and testing. I think it would be reasonable to start from `apo_continuous_action.py` and compare it only with `apo_continuous_action.py` as a test.

PR #212, opened 2022-06-21
Average PPO implementation

Howuhh:
## Description
Implementation of new algorithm - APO. See https://github.com/vwxyzjn/cleanrl/issues/210. Work in progress.

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [x] New algorithm
- [ ] Documentation

## Checklist:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ cleanrl/videos/*
 benchmark/**/*.svg
 benchmark/**/*.pkl
 mjkey.txt
+.idea
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]
--- a/benchmark/apo.sh
+++ b/benchmark/apo.sh
@@ -1,0 +1,28 @@
+export WANDB_ENTITY=openrlbenchmark
+
+poetry install -E "mujoco pybullet"
+python -c "import mujoco_py"
+
+#xvfb-run -a python -m cleanrl_utils.benchmark \
+#    --env-ids Swimmer-v3 \
+#    --command "poetry run python cleanrl/apo_continuous_action.py --track --capture-video --gae-lambda=0.99" \
+#    --num-seeds 3 \
+#    --workers 3
+
+#xvfb-run -a python -m cleanrl_utils.benchmark \
+#    --env-ids HalfCheetah-v3 \
+#    --command "poetry run python cleanrl/apo_continuous_action.py --track --capture-video --gae-lambda=0.9" \
+#    --num-seeds 3 \
+#    --workers 3
+
+#xvfb-run -a python -m cleanrl_utils.benchmark \
+#    --env-ids Ant-v3 \
+#    --command "poetry run python cleanrl/apo_continuous_action.py --track --capture-video --gae-lambda=0.9" \
+#    --num-seeds 3 \
+#    --workers 3
+
+xvfb-run -a python -m cleanrl_utils.benchmark \
+    --env-ids Hopper-v3 \
+    --command "poetry run python cleanrl/apo_continuous_action.py --track --capture-video --gae-lambda=0.99" \
+    --num-seeds 3 \
+    --workers 3
--- a/cleanrl/apo_continuous_action.py
+++ b/cleanrl/apo_continuous_action.py
@@ -1,0 +1,331 @@
+import argparse
+import os
+import random
+import time
+from distutils.util import strtobool
+
+import gym
+import numpy as np
+import pybullet_envs  # noqa
+import torch
+import torch.nn as nn
+import torch.optim as optim
+from torch.distributions.normal import Normal
+from torch.utils.tensorboard import SummaryWriter
+
+
+def parse_args():
+    # fmt: off
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--exp-name", type=str, default=os.path.basename(__file__).rstrip(".py"),
+        help="the name of this experiment")
+    parser.add_argument("--seed", type=int, default=1,
+        help="seed of the experiment")
+    parser.add_argument("--torch-deterministic", type=lambda x: bool(strtobool(x)), default=True, nargs="?", const=True,
+        help="if toggled, `torch.backends.cudnn.deterministic=False`")
+    parser.add_argument("--cuda", type=lambda x: bool(strtobool(x)), default=True, nargs="?", const=True,
+        help="if toggled, cuda will be enabled by default")
+    parser.add_argument("--track", type=lambda x: bool(strtobool(x)), default=False, nargs="?", const=True,
+        help="if toggled, this experiment will be tracked with Weights and Biases")
+    parser.add_argument("--wandb-project-name", type=str, default="cleanRL",
+        help="the wandb's project name")
+    parser.add_argument("--wandb-entity", type=str, default=None,
+        help="the entity (team) of wandb's project")
+    parser.add_argument("--capture-video", type=lambda x: bool(strtobool(x)), default=False, nargs="?", const=True,
+        help="weather to capture videos of the agent performances (check out `videos` folder)")
+
+    # Algorithm specific arguments
+    parser.add_argument("--env-id", type=str, default="HalfCheetahBulletEnv-v0",
+        help="the id of the environment")
+    parser.add_argument("--total-timesteps", type=int, default=1000000,
+        help="total timesteps of the experiments")
+    parser.add_argument("--learning-rate", type=float, default=3e-4,
+        help="the learning rate of the optimizer")
+    parser.add_argument("--num-envs", type=int, default=1,
+        help="the number of parallel game environments")
+    parser.add_argument("--num-steps", type=int, default=2048,
+        help="the number of steps to run in each environment per policy rollout")
+    parser.add_argument("--anneal-lr", type=lambda x: bool(strtobool(x)), default=True, nargs="?", const=True,
+        help="Toggle learning rate annealing for policy and value networks")
+    parser.add_argument("--hidden-dim", type=float, default=64,
+        help="Actor & Critic networks hidden dimension")
+    parser.add_argument("--gae-lambda", type=float, default=0.95,
+        help="the lambda for the general advantage estimation")
+    parser.add_argument("--num-minibatches", type=int, default=32,
+        help="the number of mini-batches")
+    parser.add_argument("--update-epochs", type=int, default=10,
+        help="the K epochs to update the policy")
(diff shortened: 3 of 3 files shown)


Howuhh:
@vwxyzjn I'd appreciate it if you could take a quick look at the code (without going into details) to check that I match the style of the rest of the library.

vwxyzjn (commented):
@Howuhh Thanks for making the PR - it looks really great. I did a pass of editing and removed some of the formatting changes -  we want to minimize the lines of code difference between say `ppo_continous_action.py` and `apo_continuous_action.py` to make it easy to spot exactly how these two algorithms are different. 

Formatting changes like `values = torch.zeros(args.num_steps, args.num_envs, device=device)` or `lr_scheduler = optim.lr_scheduler.LinearLR(optimizer, start_factor=1.0, end_factor=0.0, total_iters=num_updates)` should come in a separate PR that fixes this for all affected files, and we should also discuss the design - if we are going with the `LinearLR` route, it might be worth it to make it an argparse to configure other types of lr scheduler.

vwxyzjn:
> I also don't quite understand the decision to evaluate an agent by episodic reward and with stochastic actions. 

Are you suggesting we do deterministic evaluation by using the mean of the distribution as actions?

> This is especially noticeable with --capture-video as it slows down the training a lot (from 3k fps for 64 envs to ~700-600, 5min -> 20min for 1M). This would not be a problem if the agent was checked periodically, every few updates. This would add some code since you have to write a rollout function, but it is usually very similar for many agents.

I don't follow. What is the problem and how is it related to capture-video?

Howuhh:
> What is the problem and how is it related to capture-video

Always capturing video from one of the envs during training has a noticeable overhead (3x slower on my PC). From my experience it will be faster to not recording video during experience gathering and record it separately on single `eval_env` every N updates. However, it would require a little bit of code to create the environment, so maybe this is not a priority right now (or at all), I just wanted to share an observation.

vwxyzjn:
the video recording wrapper actually records according to a particular schedule and should not slow down training a lot. See https://www.gymlibrary.ml/content/wrappers/

vwxyzjn:
The prototype LGTM. Could you start running some benchmark experiments? 

I suggest running

```
export WANDB_ENTITY=openrlbenchmark
poetry install -E "mujoco pybullet"
python -c "import mujoco_py"
OMP_NUM_THREADS=1 xvfb-run -a python -m cleanrl_utils.benchmark \
    --env-ids HalfCheetah-v2 Walker2d-v2 Hopper-v2 \
    --command "poetry run python cleanrl/ppo_continuous_action.py --cuda False --track --capture-video" \
    --num-seeds 3 \
    --workers 9
```

Howuhh:
Yup, I started to run some stuff. At first I will be testing on Swimmer-v3 as the results on it are very different from the other environments with PPO (for the better) and so it is easier for me to compare the results in the paper.

This leaves only the question of hyperparameters. Do I have to start with the default ones? I think that some values, such as `gae-lambda` and `value-constraint` can affect the quality quite a lot.

P.S. with some hparams tweaking it learn Swimmer well ~ 360 return on 1M steps.

vwxyzjn:
Maybe leave use the same set of hyperparameters provided in the paper as a starting place. If another set of hyperparameters is used, we should clearly state this difference in the documentation.

Overall, it's desirable to have the same or better performance compared to the original paper.

Howuhh:
Well, the papper indicates only the enumeration sweep that they used, but not the final best ones. :disappointed:

vwxyzjn:
16121275-4B5D-4B2E-82A7-BAFF9F87D39C

Isn't there a hyperparameter table?

Howuhh:
Yes, it is it! The most important parameters are given here only as a grid for a sweep. It doesn't specify `num-envs`/`num-steps` also, but I suppose we could leave them as standard 1 and 2048

vwxyzjn:
Yeah let’s keep the default num_envs=1 and num_steps=2048 for now. I just looked through the original implementation and it looks like it's quite a different architecture. 

In the future we can probably try different num_envs, especially with envpool. We have some early envpool expeirments with PPO and it seems to perform really well compared to gym's vecenv.

https://wandb.ai/openrlbenchmark/envpool-cleanrl/reports/CleanRL-experiments--VmlldzoyMTA4ODM2

Howuhh:
@vwxyzjn So, what is the policy of submitting runs to the wandb? Should I first experiment on my local private project and then re-run final evaluation to the openbenchmark/cleanrl? Or I can make a project in openbenchmark/apo and submit experimantal runs here?

vwxyzjn:
Feel free to submit to openrlbenchmark/cleanrl. We don’t use WANDB_GROUP really. The grouping happens manually in the Wandb site when we create reports.

Howuhh:
First sanity-check on 3 seeds, seems like it is working as expected on Swimmer-v3. Even better than in paper, but the use more seeds.

W B Chart 22 06 2022, 17 16 16

vwxyzjn:
Feel free to either use 3 or 5 seeds. It's awesome to see that the performance is even better than the paper!

Howuhh:
@vwxyzjn Results for APO Gym Mujoco will be in this report:
https://wandb.ai/openrlbenchmark/cleanrl/reports/-WIP-APO-on-Gym-Mujoco---VmlldzoyMjEwMjY4
Feel free to edit or suggest changes.

vwxyzjn:
I modified the report a bit to make it look nicer:

image

Howuhh:
@vwxyzjn While it will be more explicit, I think we should respect the choice of the author of this method, since he called it APO. 
> try run python apo_continuous_action.py --cuda False

I don't have cuda, so it is on cpu. Without --capture-video it a lot faster tho, I can easily test one `gae-lambda` choice in 20m for 3 seeds.
Thanks for the style corrections, looks better now.

vwxyzjn:
@Howuhh are all the experiments done with the same set of hyperparameters?

Howuhh:
No, for now I vary gae-lambda as my primary goal is to replicate results from paper, not to properly compare with PPO (but I will come to that)

Howuhh:
So, the short conclusion from first experiments on Swimmer, HalfCheetah, Ant, Walker, Hopper:
1. I can match paper performance on Swimmer, HalfCheetah, Ant
2. While on Walker and Hopper performance is significantly worse for some reason.
In general pattern also match paper results: APO excels in env without termination states, while fail where they are exists (in terms of return metric, in terms of average reward APO should beat PPO in every env).

vwxyzjn:
That sounds good. Thank you!

Howuhh:
I can restore the original quality on Hopper and Walker, but with quite specific parameters. I don't think it's worth it, given that the results on these environments are only needed for pedagogical purposes (to show that the average reward is not the best choice if it's an episodic task). 

I will run benchmark on Swimmer, HalfCheetah, Ant (need to tune hparams a bit to work well on them all with same set). @vwxyzjn  Would be this enough for algorithm adoption? 

Still, I think this PR also should wait for new gym api as done bootstrapping is even more important here than in PPO. All runs I will transfer to cleanrl-cache project.

Howuhh:
> What is this? Is this an option in the gym environment?

Yeah, this option disables `done=True` on unsafe states like falling, such that done will be only on timelimit. For APO it is important as it is designed for tasks without termination states, with this option it learns on Hopper, Walker a lot quicklier and get higher reward in general. If safety is not an issue, then a don't see why we can't disable dones here. 

> It's not as clean to me if APO has a tuned parameter per game

I totally agree, I will run some tuning on all of them, it will just take some time. I will post an update with wandb report with results on all envs with same hparams.

vwxyzjn:
> Yeah, this option disables done=True on unsafe states like falling,

This changes the environment which makes a fair comparison more difficult; I am open to doing it though as long as it is clearly documented.

> I totally agree, I will run some tuning on all of them, it will just take some time. I will post an update with wandb report with results on all envs with same hparams.

Thank you! You might also be interested in a hyperparameter tuning workflow I used. 

1. Download https://wandb.ai/costa-huang/cleanRL/runs/6ai2uv8t/code?workspace=user-costa-huang and named it `tune.py`
2. Create a `sweep.yaml` file filled with
```
command:
  - ${env}
  - ${interpreter}
  - ${program}
  - ${args}
method: bayes
metric:
  goal: maximize
  name: normalized_scores
parameters:
  anneal-lr:
    values:
      - true
      - false
  clip-vloss:
    values:
      - true
      - false
  ent-coef:
    values:
      - 0
      - 0.01
  learning-rate:
    distribution: uniform
    max: 0.003
    min: 0.0003
  max-grad-norm:
    distribution: uniform
    max: 4
    min: 0.5
  num-minibatches:
    values:
      - 1
      - 2
      - 4
  num-steps:
    values:
      - 5
      - 16
      - 32
      - 64
      - 128
  update-epochs:
    values:
      - 1
      - 2
      - 4
  vf-coef:
    distribution: uniform
    max: 4
    min: 0
program: tuner.py
```
3. run `wandb sweep sweep.yaml -p cleanrl` which will prompt you to execute something like `wandb agent costa-huang/cleanRL/ngpwrthg`.

It will generate a sweep that looks like this https://wandb.ai/costa-huang/cleanRL/sweeps/ngpwrthg?workspace=user-costa-huang.

vwxyzjn:
Please disregard the hyperparameter optimization guide above and follow the https://github.com/vwxyzjn/cleanrl/pull/228 if you want to try it. 

The following is the one I used recently to tune ppo in mujoco. 


```
import optuna

from cleanrl_utils.tuner import Tuner

tuner = Tuner(
    script="cleanrl/ppo_continuous_action_envpool_jax.py",
    metric="charts/episodic_return",
    metric_last_n_average_window=50,
    direction="maximize",
    target_scores={
        "HalfCheetah-v4": [-1000, 8000],
        "Walker2d-v4": [-1000, 6000],
        "Ant-v4": [-1000, 6000],
    },
    params_fn=lambda trial: {
        "learning-rate": trial.suggest_loguniform("learning-rate", 0.0003, 0.003),
        "num-minibatches": trial.suggest_categorical("num-minibatches", [1, 2, 4]),
        "update-epochs": trial.suggest_categorical("update-epochs", [1, 2, 4]),
        "num-steps": trial.suggest_categorical("num-steps", [5, 16, 32, 64, 128]),
        "vf-coef": trial.suggest_uniform("vf-coef", 0, 5),
        "max-grad-norm": trial.suggest_uniform("max-grad-norm", 0, 5),
        "num-envs": 64
    },
    pruner=optuna.pruners.MedianPruner(n_startup_trials=5),
    wandb_kwargs={"project": "cleanrl"},
)
tuner.tune(
    num_trials=100,
    num_seeds=3,
)
```

Issue #221, opened 2022-06-27
JAX + C51


vwxyzjn:
@kinalmehta has expressed interest in working on this.

PR #224, opened 2022-06-29
Jax c51 contrib

kinalmehta:
## Description

JAX implementation for C51
Implementation for #221 

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [x] New algorithm
- [x] Documentation

## Checklist:


kinalmehta:
Results on classical gym environments can be checked here.
https://wandb.ai/kinalmehta/jax-cleanrl/reports/C51-JAX-vs-Pytorch-on-Classical-Gym-Environments--VmlldzoyNDQ3OTk5

We see a speed-up of about ~30% in the JAX version compared to Pytorch.

kinalmehta:
Here is the benchmark report on atari environments
https://wandb.ai/kinalmehta/jax-cleanrl/reports/C51-JAX-vs-Pytorch-on-Atari-Environments--VmlldzoyNjkyNDY0

**Important observations:**
- BeamRider performance is bad compared PyTorch version
- Breakout performance almost matches PyTorch variant but is still a bit low comparatively
- For Pong, the performance matches perfectly for 2 seeds, but the reward remains zero for one of the seeds.

Need to look into more detail about the differences between PyTorch and JAX implementations to fix the above mentioned issues.

joaogui1:
How does it compare to Dopamine's version?

kinalmehta:
> How does it compare to Dopamine's version?
I haven't checked Dopamine yet. I will have a look and update here, though it might take some time.

vwxyzjn:
FYI dopamine has a [benchmark](https://google.github.io/dopamine/baselines/atari/plots.html), but its x-axis is not the environment steps... Any clue on how we can compare those results? @joaogui1 
image

kinalmehta:
### Atari Fixed
After months of procrastination and debugging various aspects, I finally stumbled upon the cause of performance degradation.
The incorrect epsilon value caused this performance degradation. I missed this detail and used the default value $10^{-8}$ from optax.
However, the C51-PyTorch version uses ${0.01}/{batch\\_size}$. Hoowever I couldn't find any motivation for using this value.

Reading up more on this led to the conclusion that this is a common issue even in NLP and CV as well.
More about this hyperparameter can be read [here](https://stackoverflow.com/questions/43221065/how-does-the-epsilon-hyperparameter-affect-tf-train-adamoptimizer).

### Benchmarking classical envs on CPU
I have updated the plots of classical gym environments (CartPole, Acrobot, MountainCar) by benchmarking on CPU. We see significant speed-up compared to pytorch version on CPU.

### Comparison with dopamine
Based on the beamrider plot shared above, the below table summarizes the final score comparison
| **implementation** | **score** |
| - | - |
| dopamine | 5000-7000 |
| cleanrl-pytorch | ~9500 |
| cleanrl-jax-old | ~2500 |
| cleanrl-jax-fixed | ~9500 |

### Reports link
- https://wandb.ai/kinalmehta/cleanrl/reports/Regression-Report-c51_atari_jax--VmlldzozMjM1Mzc1
- https://wandb.ai/kinalmehta/cleanrl/reports/Regression-Report-c51_jax--VmlldzozMjM1MzM0

### Conclusion
The updated plots are available on the above links itself.
The PR looks good to be mearged once the documentation is updated. Anything else I am missing here @vwxyzjn?

vwxyzjn:
The results look incredible. Great job @kinalmehta. Thanks for chasing down the cause for the issue. The code also look great to me. Feel free to start adding documentation. You should also move the experiments to the `openrlbenchmark/cleanrl` namespace.

kinalmehta:
I've added the documentation, and now I believe this PR is ready for the final review.

docs/rl-algorithms/c51.md
```diff
@@ -16,7 +16,8 @@ Original papers:
 | ----------- | ----------- |
 | :material-github: [`c51_atari.py`](https://github.com/vwxyzjn/cleanrl/blob/master/cleanrl/c51_atari.py), :material-file-document: [docs](/rl-algorithms/c51/#c51_ataripy) |  For playing Atari games. It uses convolutional layers and common atari-based pre-processing techniques. |
 | :material-github: [`c51.py`](https://github.com/vwxyzjn/cleanrl/blob/master/cleanrl/c51.py), :material-file-document: [docs](/rl-algorithms/c51/#c51py) | For classic control tasks like `CartPole-v1`. |
-
+| :material-github: [`c51_atari.py`](https://github.com/vwxyzjn/cleanrl/blob/master/cleanrl/c51_atari_jax.py), :material-file-document: [docs](/rl-algorithms/c51/#c51_atari_jaxpy) |  For playing Atari games. It uses convolutional layers and common atari-based pre-processing techniques. |
+| :material-github: [`c51.py`](https://github.com/vwxyzjn/cleanrl/blob/master/cleanrl/c51_jax.py), :material-file-document: [docs](/rl-algorithms/c51/#c51_jaxpy) | For classic control tasks like `CartPole-v1`. |
```
vwxyzjn, line 20: These need to be renamed to c51_atari_jax.py and c51_jax.py.

The overview.md needs to be updated as well.

docs/rl-algorithms/c51.md
```diff
@@ -180,6 +181,129 @@ Tracked experiments and game play videos:
...
+<div class="grid-container">
+<img src="../c51/jax/BeamRiderNoFrameskip-v4.png">
+<img src="../c51/jax/BeamRiderNoFrameskip-v4-time.png">
+
+<img src="../c51/jax/BreakoutNoFrameskip-v4.png">
+<img src="../c51/jax/BreakoutNoFrameskip-v4-time.png">
+
+<img src="../c51/jax/PongNoFrameskip-v4.png">
+<img src="../c51/jax/PongNoFrameskip-v4-time.png">
+</div>
```
vwxyzjn, line 238: These look great! A minor note for the future: we shouldn't feel obligated to export the wandb curves manually in the future anymore -- the rlops utility should give us the `compare.png` and `compare-time.png` that we can use directly to save manual labor :)

vwxyzjn (approved):
Some minor comments. Great work! Feel free to merge once they are addressed.

One quick note is that https://github.com/vwxyzjn/cleanrl/pull/292 is getting merged. Would you mind submitting a PR to https://github.com/vwxyzjn/cleanrl/pull/292, adding model eval and uploading trained models to huggingface (for just one random seed)?

docs/rl-algorithms/c51.md
kinalmehta: done

Issue #207, opened 2022-06-20
Removing the regular advantage calculation in PPO


vwxyzjn:
# Problem description. 

The regular advantage calculation in PPO is a special case of the GAE advantage calculation when `gae_lambda=1` - we empirically demonstrate this with the debugging output in the bottom. Based on this result, we should remove 

https://github.com/vwxyzjn/cleanrl/blob/94a685de9290435623d7cf5e4e770418ddb10a4f/cleanrl/ppo.py#L232-L242

## Debugging output

```python
(cleanrl-ghSZGHE3-py3.9) ➜  cleanrl git:(explain-non-modular) ✗ ipython -i ppo.py  
Python 3.9.5 (default, Jul 19 2021, 13:27:26) 
Type 'copyright', 'credits' or 'license' for more information
IPython 8.4.0 -- An enhanced Interactive Python. Type '?' for help.
/home/costa/.cache/pypoetry/virtualenvs/cleanrl-ghSZGHE3-py3.9/lib/python3.9/site-packages/gym/utils/passive_env_checker.py:97: UserWarning: WARN: We recommend you to use a symmetric and normalized Box action space (range=[-1, 1]) https://stable-baselines3.readthedocs.io/en/master/guide/rl_tips.html
  logger.warn(
/home/costa/.cache/pypoetry/virtualenvs/cleanrl-ghSZGHE3-py3.9/lib/python3.9/site-packages/gym/core.py:200: DeprecationWarning: WARN: Function `env.seed(seed)` is marked as deprecated and will be removed in the future. Please use `env.reset(seed=seed)` instead.
  deprecation(
global_step=36, episodic_return=9.0
global_step=52, episodic_return=13.0
global_step=100, episodic_return=25.0
global_step=112, episodic_return=19.0
global_step=128, episodic_return=32.0
global_step=144, episodic_return=11.0
global_step=152, episodic_return=13.0
global_step=176, episodic_return=12.0
global_step=196, episodic_return=11.0
global_step=228, episodic_return=13.0
global_step=260, episodic_return=16.0
global_step=296, episodic_return=46.0
global_step=300, episodic_return=39.0
global_step=312, episodic_return=13.0
global_step=360, episodic_return=15.0
global_step=388, episodic_return=40.0
global_step=400, episodic_return=22.0
global_step=408, episodic_return=28.0
global_step=440, episodic_return=13.0
global_step=460, episodic_return=15.0
global_step=484, episodic_return=31.0
[...]

PR #234, opened 2022-07-19
Implement PPO-DNA algorithm for Atari

jseppanen:
## Description

Add implementation of PPO-DNA algorithm for Atari Envpool.

### Paper reproduction (attempt)

Here's the episodic rewards after 200M environment steps (50M gradient updates), compared to Fig. 6 in the original paper:
* BattleZone: 121 000 ± 26 000 (better than about 60 000 in the paper)
* DoubleDunk: -3.8 ± 0.4 (worse than about 1.0 in the paper)
* NameThisGame: 10 700 ± 4 300 (worse than about 20 000 in the paper)
* Phoenix: 208 000 ± 73 000 (better than about 80 000 in the paper)
* Qbert: 12 000 ± 5 000 (worse than about 30 000 in the paper)

However, I used the default networks and environments from the CleanRL PPO Atari implementation, so there are probably differences in them, vs. the original paper. In summary, when comparing against paper results, this implementation gets better returns in two out of the five tasks.

Results from figure 6 in the paper:
figure6

When comparing against ClearRL PPO Atari Envpool implementation, this implementation performs better on six out of nine tasks. See detailed learning curves, compared against CleanRL PPO Atari Envpool implementation:
[PPO-DNA vs PPO on Atari Envpool](https://wandb.ai/jseppanen/cleanrl/reports/PPO-DNA-vs-PPO-on-Atari-Envpool--VmlldzoyMzM5Mjcw)

### Reference

* [DNA: Proximal Policy Optimization with a Dual Network Architecture](https://arxiv.org/abs/2206.10027)

cc @maitchison

## Types of changes
- [ ] Bug fix
- [ ] New feature
- [x] New algorithm
- [x] Documentation

## Checklist:

- [x ] I've read the [CONTRIBUTION](https://github.com/vwxyzjn/cleanrl/blob/master/CONTRIBUTING.md) guide (**required**).
- [x ] I have ensured `pre-commit run --all-files` passes (**required**).
- [x ] I have updated the documentation.

    - [x ] I have explained note-worthy implementation details.
    - [x ] I have explained the logged metrics.
    - [x ] I have added links to the original paper and related papers (if applicable).
    - [x ] I have created a table comparing my results against those from reputable sources (i.e., the original paper or other reference implementation).
    - [x ] I have added the learning curves (in PNG format with `width=500` and `height=300`).
    - [x ] I have added links to the tracked experiments.


cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,474 @@
...
+        help="total timesteps of the experiments")
+    parser.add_argument("--num-envs", type=int, default=128,
+        help="the number of parallel game environments")
+    parser.add_argument("--envpool-num-threads", type=int, default=2,
+        help="the number of envpool worker threads")
+    parser.add_argument("--num-steps", type=int, default=128,
+        help="rollout horizon: the number of steps to run in each environment per policy rollout")
+    parser.add_argument("--anneal-lr", type=lambda x: bool(strtobool(x)), default=True, nargs="?", const=True,
+        help="Toggle learning rate annealing for policy and value networks")
+    parser.add_argument("--gamma", type=float, default=0.99,
```
vwxyzjn, line 53: Note a complete review yet, just putting a note here: the author suggests the `gamma` parameter used in DNA should be 0.999. I am ok with using either 0.99 as long as this difference is spelled out.

maitchison:
Small thing  

> Here's the episodic rewards after 200M environment steps (50M gradient updates), compared to Fig. 6 in the original paper:

should be 

> Here's the episodic rewards after 200M environment steps (50M environment interactions), compared to Fig. 6 in the original paper:

The algorithm will make many more than 50M gradient updates due to the number of mini-batches.

cleanrl/ppo_dna_atari_envpool.py
maitchison: I'd recommend using the same settings as the paper, which would be the following changes to the defaults

gamma=0.999 
clip_coef=0.2
anneal-lr=false 
max_grad_norm=5.0

Gamma and clip_coef  make a fairly big difference too. The other settings (max_grad_norm, anneal etc) are not so important.

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,474 @@
...
+    # env setup
+    envs = envpool.make(
+        args.env_id,
+        env_type="gym",
+        num_envs=args.num_envs,
+        # batch_size=args.epoch_size,  # async
+        num_threads=args.envpool_num_threads,
+        episodic_life=True,
+        reward_clip=True,
+        seed=args.seed,
```
maitchison, line 260: Just a note. We designed DNA to run on the original rewards, and it works best with reward_clip off.  That being said, this would mean enabling reward normalization (i.e. dividing by the variance of returns).

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,474 @@
...
+                value_optimizer.zero_grad()
+                v_loss.backward()
+                nn.utils.clip_grad_norm_(agent_value.parameters(), args.max_grad_norm)
+                value_optimizer.step()
+
+        # Value network to policy network distillation
+        agent_policy.zero_grad(True)  # don't clone gradients
+        old_agent_policy = deepcopy(agent_policy)
+        old_agent_policy.eval()
+        for epoch in range(args.distill_update_epochs):
```
maitchison, line 419: Just a note, in future work we found by switching the order from Rollout-Value-Policy-Distil, to Rollout-Distil-Policy-Value we could avoid an additional forward (and the copy you need here too) and instead simply use the policy that was generated during rollout. This means the policy updates are slightly off-policy, even so, imperically it works slightly better than the order used in our paper.

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,474 @@
...
+
+import envpool
+import gym
+import numpy as np
+import torch
+import torch.nn as nn
+import torch.optim as optim
+from torch.distributions import Categorical, kl_divergence
+from torch.utils.tensorboard import SummaryWriter
+
```
maitchison, line 20: You'll probably need to add 

> torch.backends.cuda.matmul.allow_tf32 = False

before the first torch operation, otherwise, you may have issues with 30xx series cards. This issue has been fixed in torch 1.12, but I see we are pinned to an earlier version. Without it, some operations will be reduced to 10-bits of precision (around 3 sig fig) which may cause problems with, for example, the observation normalization. See (https://pytorch.org/docs/stable/notes/cuda.html#tf32-on-ampere) for a discussion on this issue.

cleanrl/ppo_dna_atari_envpool.py
jseppanen: yeah, I was originally thinking to have minimal changes from vanilla PPO hyperparams, but I agree it's better to have well tuned algorithm specific defaults out of the box.

jseppanen: I'm also updating entropy coef, as it's 0.001 in the paper, vs. 0.01 in baseline PPO

cleanrl/ppo_dna_atari_envpool.py
jseppanen: yeah I was actually thinking about the same thing, and maybe that could explain (part of) the differences in the atari-5 repro results.

cleanrl/ppo_dna_atari_envpool.py
jseppanen: sounds great, I'll flip the order!

maitchison:
@jseppanen This latest change looks great. It seems, however, that I made a typo in the paper. The entropy bonus should be 1e-2 (0.01) not 0.001 as I have mistakenly written in the paper. I think 1e-3 will likely cause the algorithm to underperform the default 0.01 setting. I've made a note of this and will fix it in the next revision. Thanks for spotting this!

cleanrl/ppo_dna_atari_envpool.py
jseppanen: I removed the clipping and added reward normalization

jseppanen:
@maitchison OK I addressed all of your comments, plus added one extra change (used builtin `gym.wrappers.NormalizeObservation` instead of DIY) – I'm still going to re-run some of the experiments to verify that it still works 😅

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,468 @@
...
+
+def parse_args():
+    # fmt: off
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--exp-name", type=str, default=os.path.basename(__file__).rstrip(".py"),
+        help="the name of this experiment")
+    parser.add_argument("--seed", type=int, default=np.random.randint(2 ** 31),
```
vwxyzjn, line 26: Please revert the default seed to 1. This is an expected behavior of using CleanRL that the seeds by default is the same for reproducibility and debugging reason.

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,468 @@
...
+        help="if toggled, this experiment will be tracked with Weights and Biases")
+    parser.add_argument("--wandb-project-name", type=str, default="cleanRL",
+        help="the wandb's project name")
+    parser.add_argument("--wandb-entity", type=str, default=None,
+        help="the entity (team) of wandb's project")
+
+    # Algorithm specific arguments
+    parser.add_argument("--env-id", type=str, default="Pong-v5",
+        help="the id of the environment")
+    parser.add_argument("--total-timesteps", type=int, default=10_000_000,
```
vwxyzjn, line 42: Here we use the style `10000000`. I'd be open to using `10_000_000`, but this change should apply to all of our variants in the form of another PR.

 It's probably best to automate this using something like a pre-commit hook that automatically converts all instances of `10000000` to `10_000_000`.

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,468 @@
...
+        help="total timesteps of the experiments")
+    parser.add_argument("--num-envs", type=int, default=128,
+        help="the number of parallel game environments")
+    parser.add_argument("--envpool-num-threads", type=int, default=2,
```
vwxyzjn, line 46: This change should also apply to all envpool scripts in another PR.

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,468 @@
...
+        help="the number of envpool worker threads")
+    parser.add_argument("--num-steps", type=int, default=128,
+        help="rollout horizon: the number of steps to run in each environment per policy rollout")
```
vwxyzjn, line 49: This "rollout horizon: " change should also apply to all PPO scripts in another PR.

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,468 @@
...
+    parser.add_argument("--anneal-lr", type=lambda x: bool(strtobool(x)), default=False, nargs="?", const=True,
+        help="Toggle learning rate annealing for policy and value networks")
+    parser.add_argument("--gamma", type=float, default=0.999,
+        help="the discount factor gamma")
+
+    parser.add_argument("--policy-learning-rate", type=float, default=2.5e-4,
+        help="the learning rate of the policy optimizer")
+    parser.add_argument("--policy-batch-size", type=int, default=2048,
```
vwxyzjn, line 57: put a note here

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,468 @@
...
+        returns = torch.zeros_like(rewards)
+        for t in reversed(range(total_steps)):
+            if t == total_steps - 1:
+                nextnonterminal = 1.0 - next_done
+                next_return = next_value
+            else:
+                nextnonterminal = 1.0 - dones[t + 1]
+                next_return = returns[t + 1]
+            returns[t] = rewards[t] + gamma * nextnonterminal * next_return
+        advantages = returns - values
```
vwxyzjn, line 169: This can be removed. See https://github.com/vwxyzjn/cleanrl/issues/207

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,468 @@
...
+            nn.ReLU(),
+            nn.Flatten(),
+            layer_init(nn.Linear(64 * 7 * 7, 512)),
+            nn.ReLU(),
+        )
+        self.actor = layer_init(nn.Linear(512, envs.single_action_space.n), std=0.01)
+        self.critic = layer_init(nn.Linear(512, 1), std=1)
+
+    def forward(self, x):
+        """Run policy and value functions."""
```
vwxyzjn, line 197: This comment is not necessary.

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,468 @@
...
+    @staticmethod
+    def sample_action(logits):
+        probs = Categorical(logits=logits)
+        action = probs.sample()
+        return action, probs.log_prob(action)
+
+    @staticmethod
+    def action_log_prob(logits, action):
+        probs = Categorical(logits=logits)
+        return probs.log_prob(action), probs.entropy()
```
vwxyzjn, line 212: Please keep this part of the code as similar to the following as you can.

```
    def get_value(self, x):
        return self.critic(self.network(x / 255.0))

    def get_action_and_value(self, x, action=None):
        hidden = self.network(x / 255.0)
        logits = self.actor(hidden)
        if action is None:
            action = probs.sample()
        return action, probs.log_prob(action), probs.entropy(), self.critic(hidden)
```

To still make it work w/ the distillation loss, maybe you can do

`return action, probs.log_prob(action), probs.entropy(), self.critic(hidden), logits`

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,468 @@
...
+        )
+    writer = SummaryWriter(f"runs/{run_name}")
+    writer.add_text(
+        "hyperparameters",
+        "|param|value|\n|-|-|\n%s" % ("\n".join([f"|{key}|{value}|" for key, value in vars(args).items()])),
+    )
+
+    # Fix multiplication precision on Ampere devices (30xx) prior to PyTorch 1.12
+    # https://pytorch.org/docs/stable/notes/cuda.html#tf32-on-ampere
+    torch.backends.cuda.matmul.allow_tf32 = False
```
vwxyzjn, line 244: Could you do some preliminary experiments to check if this makes a difference? It's important that we understand the effect of the techniques added.

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,468 @@
...
+        episodic_life=True,
+        seed=args.seed,
+    )
+    envs.is_vector_env = True
+    envs.num_envs = args.num_envs
+    envs.single_action_space = envs.action_space
+    envs.single_observation_space = envs.observation_space
+    envs = RecordEpisodeStatistics(envs)
+    # if args.capture_video:
+    #     envs = gym.wrappers.RecordVideo(envs, f"videos/{run_name}")
```
vwxyzjn, line 270: Great catch. Please remove this.

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,468 @@
...
+    envs = gym.wrappers.NormalizeObservation(envs)
+    envs = gym.wrappers.TransformObservation(envs, lambda obs: np.clip(obs, -3, 3))
```
vwxyzjn, line 272: I am a little hesitant to use `gym.wrappers.NormalizeObservation(envs)` in image-based envs. Dividing the pixels by 255 is a well-known normalization technique. `ppo_procgen.py`, for example, uses only the `gym.wrappers.NormalizeReward`.

https://github.com/vwxyzjn/cleanrl/blob/c41bee932ef998fcbbe301c6d91c2328c1c03ab7/cleanrl/ppo_procgen.py#L188-L197

To this end, my suggestion is not to use wrapper if it does not result in performance benefits.

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,468 @@
...
+        return probs.log_prob(action), probs.entropy()
+
+    @staticmethod
+    def action_kl_divergence(old_logits, new_logits):
+        old_probs = Categorical(logits=old_logits)
+        new_probs = Categorical(logits=new_logits)
+        return kl_divergence(old_probs, new_probs).mean()
+
+
+def main():
```
vwxyzjn, line 221: Out of curiosity, what is the purpose of using the `main()` function? We loss a bit of flexibility in debug when we do something like 

```
python ppo_dna_atari_envpool.py

# run it for 10 seconds then ctrl + c to print variables
```

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,468 @@
...
+    agent_policy = Agent(envs).to(device)
+    agent_value = Agent(envs).to(device)
+    policy_optimizer = optim.Adam(agent_policy.parameters(), lr=args.policy_learning_rate, eps=1e-5)
+    value_optimizer = optim.Adam(agent_value.parameters(), lr=args.value_learning_rate, eps=1e-5)
+    distill_optimizer = optim.Adam(agent_policy.parameters(), lr=args.distill_learning_rate, eps=1e-5)
+
+    if args.track:
+        # log gradients
+        wandb.watch(agent_policy)
+        wandb.watch(agent_value)
```
vwxyzjn, line 286: Is gradient logging useful for debugging?

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,468 @@
...
+    actions = torch.zeros((args.num_steps, args.num_envs) + envs.single_action_space.shape).to(device)
+    logprobs = torch.zeros((args.num_steps, args.num_envs)).to(device)
+    rewards = torch.zeros((args.num_steps, args.num_envs)).to(device)
+    dones = torch.zeros((args.num_steps, args.num_envs)).to(device)
+    values = torch.zeros((args.num_steps, args.num_envs)).to(device)
+    avg_returns = deque(maxlen=20)
+
+    # TRY NOT TO MODIFY: start the game
+    global_step = 0
+    env_step = 0
```
vwxyzjn, line 300: Not necessary. We know the frame skip number already, and we can post-process this.

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,468 @@
...
+            action_logits[step] = logit
+            actions[step] = action
+            logprobs[step] = logprob
+            values[step] = value
+
+            # TRY NOT TO MODIFY: execute the game and log data.
+            next_obs, reward, done, info = envs.step(action.cpu().numpy())
+            rewards[step] = torch.tensor(reward, dtype=torch.float32).to(device).view(-1)
+            next_obs = torch.tensor(next_obs, dtype=torch.float32).to(device)
+            next_done = torch.tensor(done, dtype=torch.float32).to(device)
```
vwxyzjn, line 334: This difference should be kept to a minimum.

image

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,468 @@
...
+
+    envs.close()
+    writer.close()
+
+    if args.track:
+        # save final model
+        torch.save(agent_policy.state_dict(), "agent_policy.pt")
+        torch.save(agent_value.state_dict(), "agent_value.pt")
+        wandb.save("agent_policy.pt")
+        wandb.save("agent_value.pt")
```
vwxyzjn, line 464: Out of curiosity, what is the use case of this?

vwxyzjn (changes requested):
Thank you so much for your contribution. The code has great readability, and thank you for including the documentation and benchmark results.

Most of my comments are related to style - generally speaking, we try to minimize the lines of different code across files. Such a practice has two benefits
1. the filediff clearly shows the necessary changes for implementing another algorithm,
2. it makes our single-file codebase more maintainable.

To this end, please try to keep the original structure as much as possible, and I suggest walking through the code difference between `ppo_atari_envpool.py` and `ppo_dna_atari_envpool.py` and remove any unnecessary changes.


image

cleanrl/ppo_dna_atari_envpool.py
jseppanen: 👍

cleanrl/ppo_dna_atari_envpool.py
jseppanen: yeah it's sometimes been useful, esp. when trying to choose the gradient clipping value, but I can remove this to be in line with the baseline PPO style

cleanrl/ppo_dna_atari_envpool.py
jseppanen: yeah, it's because I extracted `compute_advantages` as a separate function, but then noticed that it was using values of global variables defined later in the module, which is not what I intended. It's essentially a safeguard to remove all global variables with generic names like `reward` or `value` etc.

cleanrl/ppo_dna_atari_envpool.py
jseppanen: 👍

cleanrl/ppo_dna_atari_envpool.py
jseppanen: 👍

cleanrl/ppo_dna_atari_envpool.py
jseppanen: I think I'll just revert this

cleanrl/ppo_dna_atari_envpool.py
jseppanen: I don't follow, you mean a comment explaining why there is an empty line? or something else?

cleanrl/ppo_dna_atari_envpool.py
jseppanen: but it looks like that code is still there in both `ppo_atari.py` and `ppo_atari_envpool.py` as well – maybe I'll keep it in for now, and then later all three files can be changed with a dedicated PR, wdyt?

cleanrl/ppo_dna_atari_envpool.py
jseppanen: 👍

cleanrl/ppo_dna_atari_envpool.py
jseppanen: unfortunately I don't have access to Ampere devices, so maybe I'll revert this as well

cleanrl/ppo_dna_atari_envpool.py
jseppanen: 👍

cleanrl/ppo_dna_atari_envpool.py
jseppanen: well this is meant to reproduce the DNA paper, where they suggest doing exactly that (normalize images to zero mean unity variance, then clip to -3...3)

should I add a comment to explain why, or no?

cleanrl/ppo_dna_atari_envpool.py
jseppanen: removed

cleanrl/ppo_dna_atari_envpool.py
jseppanen: OK yeah, I'll drop it from here

cleanrl/ppo_dna_atari_envpool.py
jseppanen: Of course, then I fixed the compute_advantages arguments list to include all used variables, so it's no longer using global variables (I think), but in general, I feel it would be nicer that the program fails with an undefined variable instead of silently using a global variable defined somewhere else

However, I realize that this is not a problem if everything is in global scope and there are no functions (which pretty much is the cleanRL style), but I think that the compute_advantages function is necessary in this case, and that's why I opted for increased safety against (future) bugs, at the cost of decreased interactive convenience

Personally, I edit the file locally to replace `def main():` with `if 1:` for a smooth interactive experience, if needed. But I'm happy to revert this change as well if you think so; lmk!

cleanrl/ppo_dna_atari_envpool.py
jseppanen: done

cleanrl/ppo_dna_atari_envpool.py
jseppanen: done

jseppanen:
OK so some news:

I reverted the hyperparameter change (commit 0fe7b1f) and the learning stage ordering change (commit e50bf9e). The reason is that I did some investigation to what changes led to the algorithm not being able to solve Pong in 3M steps. What I found out was that:
1. using gym observation normalization (commit cde9ada) and removing reward clipping & adding reward normalization (commit 7bd5244) alone were good, no performance regressions there
2. however, when I applied the hyperparameter change (commit 0fe7b1f) on top of observation & reward normalization, then it could suddenly no longer learn Pong
3. to cross-check, I also tested only applying the new hyperparameters without any other changes from original version (commit 981201f), and although it was able to start learning Pong, there was a reduction in learning speed/sample efficiency
I didn't want to start digging into which specific hyperparameter was triggering the problems but simply reverted the whole change for now.

Then I had also noticed another regression in wall clock learning speed, and also, because @vwxyzjn encouraged me to try and keep the code changes as small as possible, compared to the baseline `ppo_atari_envpool.py` algorithm, I also decided to revert the learning stage order change (commit e50bf9e). Maybe that could be a separate PR one day, with separate experiments to back it up.

Finally, I went through the whole file and reverted unnecessary differences compared to `ppo_atari_envpool.py` and this brings us to the current commit, 94fc331. Based on quick Pong experiment, it's working as well as the original version, so I'll be launching hopefully final experiments to make sure that it's still working like the original implementation.

The hyperparameters are a bit of a mystery, and it would be nice to get more visibility on what's going on there, but I'm afraid that would require some hyperparameters sweeping on a good set of Atari games, and I'm not sure I'm in the mood for that right now 😅

Dipamc77:
Code looks good, matches details mentioned in the paper. Really like the idea that just using a different lambda for advantage estimation gives such a huge boost overall.

One important difference seems that PPO-DNA consistently uses 128 parallel envs and PPO by default is using 8. I'm not aware if using higher makes too much of a difference in PPO but selecting 128 for PPO-DNA seems to be a conscious choice (paper does give some explanation on this as well)

Side note, the paper cites PPG as the closest work but I'd argue this is closer [Decoupling Value and Policy for Generalization in Reinforcement Learning](https://arxiv.org/pdf/2102.10330.pdf)

cleanrl/ppo_dna_atari_envpool.py
Dipamc77: If we'd like to have all the variables globally, its possible to add two extra lines in the original GAE calculation code to have the two separate advantage calculation steps. I think almost everything else remains same.

```
delta_policy = rewards[t] + gamma * nextvalues * nextnonterminal - values[t]
advantages_policy[t] = lastgaelam = delta_policy + gamma * policy_lambda * nextnonterminal * lastgaelam
delta_value = rewards[t] + gamma * nextvalues * nextnonterminal - values[t]
advantages_value[t] = lastgaelam = delta_value + gamma * value_lambda * nextnonterminal * lastgaelam
``` 

Generally I prefer functions more but cleanrl avoids them so this can be an alternative.

vwxyzjn: Thanks @Dipamc77 for the suggestion. In this case though, using `compute_advantages` seems pretty clean to me, so I am ok with it :)

vwxyzjn: >  It's essentially a safeguard to remove all global variables with generic names like reward or value etc.

This is a good point. I was thinking of using a pre-commit hook like pylint to detect this kind of situations. For now, let's revert back to not using the `main()` function to be consistent w/ other scripts

cleanrl/ppo_dna_atari_envpool.py
vwxyzjn: Sorry, that sentence was mainly for me. My original thought was that the other PPO related scripts have used `--num-minibatches`, so in this case we might be able to change it to `--num-policy-minibatches`. I will let you decide on this though. Feel free to pick one that feels like the cleanest design to you.

cleanrl/ppo_dna_atari_envpool.py
vwxyzjn: If the paper specifically normalizes like this way then it's fine.

cleanrl/ppo_dna_atari_envpool.py
vwxyzjn: Either way works for me since #207 is a planned PR. If you decide to leave it in this PR, then #207 will take care of it :)

cleanrl/ppo_dna_atari_envpool.py
vwxyzjn: I have ampere devices - once things are finalized I will be happy to give it a try.

cleanrl/ppo_dna_atari_envpool.py
vwxyzjn: Interesting - could you elaborate on the process of choosing the gradient clipping value?

cleanrl/ppo_dna_atari_envpool.py
vwxyzjn: How would you generate gameplay videos, though? Through the gym wrappers? This might be a legitimate use case for model saving.

vwxyzjn:
> PPO-DNA consistently uses 128 parallel envs

This kind of changes things a lot. I am more inclined to reproduce this work similar to how the paper is done. Please try to use 128 parallel envs if possible.

cleanrl/ppo_dna_atari_envpool.py
jseppanen: Ah true; I changed the config to use minibatch size, to be able to easily match/verify the values against the the paper

cleanrl/ppo_dna_atari_envpool.py
jseppanen: done

cleanrl/ppo_dna_atari_envpool.py
jseppanen: well I just eyeball the gradient norms over the course of training and compare against returns at the same time, and try to guess if SGD updates might be overshooting

jseppanen:
Update: I re-ran the experiments with commit 94fc331 and updated the documentation/results, and removed the main() function.

@vwxyzjn github is telling me that "requested changes must be addressed" but I can't find which changes are still needed, sorry for that. I think I addressed all your comments but is there something still that I missed?

jseppanen:
> One important difference seems that PPO-DNA consistently uses 128 parallel envs and PPO by default is using 8. I'm not aware if using higher makes too much of a difference in PPO but selecting 128 for PPO-DNA seems to be a conscious choice (paper does give some explanation on this as well)

Yeah I noticed a big improvement when using 128 envs, so all the experiments were done with that.

maitchison:
> 

The large number of paralell agents helps a lot with dealing the gradient noise. Parallel agents are better than longer n_steps due to the n_steps producing temporally correlated trajectories. One of the big tricks to getting PG to work well is using *a lot* of parallel agents. From memory the Hide-and-seek paper (https://arxiv.org/abs/1909.07528) used batchs of tens-of-thousands. ProcGen uses 64k. I've tried higher than 128, but gains diminish quickly after that on Atari, and memory becomes an issue. 

A large number of Parallel is also compute efficient as forwarding the model and parallelizing the environments all become faster.

maitchison:
> Code looks good, matches details mentioned in the paper. Really like the idea that just using a different lambda for advantage estimation gives such a huge boost overall.
> 
> One important difference seems that PPO-DNA consistently uses 128 parallel envs and PPO by default is using 8. I'm not aware if using higher makes too much of a difference in PPO but selecting 128 for PPO-DNA seems to be a conscious choice (paper does give some explanation on this as well)
> 
> Side note, the paper cites PPG as the closest work but I'd argue this is closer [Decoupling Value and Policy for Generalization in Reinforcement Learning](https://arxiv.org/pdf/2102.10330.pdf)

Good spotting! We are actually in the process now of updating the paper to discuss this reference :)

maitchison:
> OK so some news:
> 
> I reverted the hyperparameter change (commit [0fe7b1f](https://github.com/vwxyzjn/cleanrl/commit/0fe7b1f1b4d846cc6e61e5ddb1547e5664bbb6e3)) and the learning stage ordering change (commit [e50bf9e](https://github.com/vwxyzjn/cleanrl/commit/e50bf9e220360e3fc049316c1dbaa825f15b9e4b)). The reason is that I did some investigation to what changes led to the algorithm not being able to solve Pong in 3M steps. What I found out was that:
> 
> 1. using gym observation normalization (commit [cde9ada](https://github.com/vwxyzjn/cleanrl/commit/cde9adaf46629bb406e1c15c84ff565fcb5cdf6f)) and removing reward clipping & adding reward normalization (commit [7bd5244](https://github.com/vwxyzjn/cleanrl/commit/7bd52441e561168b7fdfe1f089e61f2a90144dc1)) alone were good, no performance regressions there
> 2. however, when I applied the hyperparameter change (commits [0fe7b1f](https://github.com/vwxyzjn/cleanrl/commit/0fe7b1f1b4d846cc6e61e5ddb1547e5664bbb6e3) & [1b91a46](https://github.com/vwxyzjn/cleanrl/commit/1b91a46a74f49c2507de31e931f375fad4b1cb88)) on top of observation & reward normalization, then it could suddenly no longer learn Pong
> 3. to cross-check, I also tested only applying the new hyperparameters without any other changes from original version (commit [981201f](https://github.com/vwxyzjn/cleanrl/commit/981201f39978c7162a437a256a616cd03ff1aaf4)), and although it was able to start learning Pong, there was a reduction in learning speed/sample efficiency
> 
> I didn't want to start digging into which specific hyperparameter was triggering the problems but simply reverted the whole change for now.
> 
> Then I had also noticed another regression in wall clock learning speed, and also, because @vwxyzjn encouraged me to try and keep the code changes as small as possible, compared to the baseline `ppo_atari_envpool.py` algorithm, I also decided to revert the learning stage order change (commit [e50bf9e](https://github.com/vwxyzjn/cleanrl/commit/e50bf9e220360e3fc049316c1dbaa825f15b9e4b)). Maybe that could be a separate PR one day, with separate experiments to back it up.
> 
> Finally, I went through the whole file and reverted unnecessary differences compared to `ppo_atari_envpool.py` and this brings us to the current commit, [94fc331](https://github.com/vwxyzjn/cleanrl/commit/94fc331d2ecdcbe70279e72d719787a3e75f83f1). Based on quick Pong experiment, it's working as well as the original version, so I'll be launching hopefully final experiments to make sure that it's still working like the original implementation.
> 
> The hyperparameters are a bit of a mystery, and it would be nice to get more visibility on what's going on there, but I'm afraid that would require some hyperparameters sweeping on a good set of Atari games, and I'm not sure I'm in the mood for that right now sweat_smile

I'm not sure if this is helpful, but I was reading through some of my old notes and found a note saying DNA did not converge on Pong without observation normalization. Perhaps the regression could have caused issues by the normalization? We normed to unit normal (via a running mean and second moment), then clipped to -5,+5, then multiplied by 3.0. But anything reasonable should do. We also used the default pytorch weight initialization, rather than orthogonal. I don't think that'd matter much though.

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,432 @@
...
+    # TRY NOT TO MODIFY: seeding
+    random.seed(args.seed)
+    np.random.seed(args.seed)
+    torch.manual_seed(args.seed)
+    torch.backends.cudnn.deterministic = args.torch_deterministic
+
+    device = torch.device("cuda" if torch.cuda.is_available() and args.cuda else "cpu")
+
+    # env setup
+    envs = envpool.make(
```
vwxyzjn, line 238: Did you decide not to use sticky action? See `repeat_action_probability` in envpool's documentation. Also CC @maitchison .

docs/rl-algorithms/ppo_dna.md
```diff
@@ -0,0 +1,101 @@
...
+
+
+### Experiment results
+
+Below are the average episodic returns for `ppo_dna_atari_envpool.py` compared to `ppo_atari_envpool.py`.
+
+
+| Environment      | `ppo_dna_atari_envpool.py` | `ppo_atari_envpool.py` |
+| ----------- | ----------- | ----------- | 
+| BattleZone-v5 (40M steps) | 74000 ± 15300 | 28700 ± 6300 |
```
vwxyzjn, line 60: Do you mean 50M steps?

docs/rl-algorithms/ppo_dna.md
```diff
@@ -0,0 +1,101 @@
...
+| BeamRider-v5 (10M steps) | 5200 ± 900 | 1900 ± 530 |
+| Breakout-v5 (10M steps) | 319 ± 63 | 349 ± 42 |
```
vwxyzjn, line 62: I suggest being consistent and always using 50M steps. It's ok to compare the `ppo_atari_envpool.py` that only has 10M steps. If you think comparing 50M steps is warranted, please start a `ppo_atari_envpool_50M` wandb project in the `openrlbenchmark` entity and run `ppo_atari_envpool.py` in 50M steps there.

docs/rl-algorithms/ppo_dna.md
```diff
@@ -0,0 +1,101 @@
...
+| DoubleDunk-v5 (40M steps) | -4.1 ± 1.0 | -2.0 ± 0.8 |
+| NameThisGame-v5 (40M steps) | 19100 ± 2300 | 4400 ± 1200 |
+| Phoenix-v5 (45M steps) | 186000 ± 67000 | 9900 ± 2700 |
+| Pong-v5 (3M steps) | 19.5 ± 1.0 | 16.6 ± 2.4 |
```
vwxyzjn, line 66: Same comment as above. Please run using 50M steps.

vwxyzjn (commented):
Thanks for your patience! You have addressed my comments and please ignore the `requested changes` from Github. 

The next step is to contact me to have yourself added in the `openrlbenchmark` wandb team. Then you may either move your experiments in `openrlbenchmark/cleanrl` or re-run some of them.

I added some small refactoring in the last two commits. Besides, I updated the checklist to include a few more items. 

- [ ]  I have updated the overview sections at the [docs](https://docs.cleanrl.dev/rl-algorithms/overview/) and the [repo](https://github.com/vwxyzjn/cleanrl#overview)
- [ ] I have added the commands used to run experiments in the `benchmark` folder like [benchmark/ppo.sh](https://github.com/vwxyzjn/cleanrl/blob/master/benchmark/ppo.sh)

Please also add an end-to-end test in 

https://github.com/vwxyzjn/cleanrl/blob/df3ab2a1843b906c6e528d37e1eb0a34e831e5db/tests/test_envpool.py#L4-L9

cleanrl/ppo_dna_atari_envpool.py
jseppanen: I took the config from `ppo_atari_envpool.py` which doesn't use sticky actions, and I wouldn't like to change that now, because that would unfortunately require re-running all experiments.

docs/rl-algorithms/ppo_dna.md
jseppanen: Actually, my baseline `ppo_atari_envpool.py` experiments crashed before reaching full 50M steps, but I think the results are still valuable because they quite clearly show the performance difference between CleanRL baseline PPO and PPO-DNA.

docs/rl-algorithms/ppo_dna.md
jseppanen: Sure, though I'm using rented cloud GPU's so re-running has a direct cost to me.

Also, I think there's already a couple of layers of evidence for PPO-DNA without running any more experiments:
1. Primarily, I was interested in reproducing the paper results on the five Atari envs shown in Fig 6 in the paper. The paper used 50M steps for their experiments, so that's what I did as well. My results are not exactly the same, but in the same ballpark (one env performs better, two are comparable, and two are worse).
2. Secondarily, I wanted to do an apples-to-apples comparison against the baseline `ppo_atari_envpool.py` implementation, because that way the environment settings and other implementation details wouldn't distort the results. The CleanRL documentation has learning curves for Breakout, Pong, and BeamRider for 10M steps, so that's what I ran (plus I noticed some Tennis runs in the W&B project). I attempted to run the comparisons for 50M steps as well, but the baseline runs crashed (due to my cloud account running dry), so I thought 40M steps is already more than enough.

Therefore, I'm not sure what the benefit from re-running `ppo_atari_envpool.py` for 50M steps would be?

With regards to hyperparameters, they are otherwise the same, except for total_timesteps.

docs/rl-algorithms/ppo_dna.md
jseppanen: What I could do instead is maybe show the results at the 10M steps mark, if 40M is a problem?

vwxyzjn:
Starting a new thread for discussion. @jseppanen, thanks for bearing with me. The results are definitely convincing, but the main issue is consistency (same configuration, same steps, non-crashed runs). Totally understand that running experiments can be expensive. I can help run the experiments, but it will take some time (probably a month).

jseppanen:
@vwxyzjn sure, I added benchmarks/ppo_dna.sh.  Also, maybe I haven't communicated my experiment results clearly enough, I could also try and consolidate them into one place.

bragajj:
@jseppanen wondering if you're still interested in this PR. I took a quick look at your wandb and see runs in your account for Phoenix-v5, NamesThisGame-v5, DoubleDunk-v5 at 50M steps, just not Pong. Trying to understand where we ended up here.

Was it that you were reporting results at different step counts despite having 50M step runs completed? Or were those runs not completed successfully? Just want to see if I can provide any help or guidance.

maitchison:
Hi @vwxyzjn, I now have some time I can put into this and would be happy to finish off the last few things that need doing. Looks like @jseppanen has got it mostly there, so it shouldn't take too long. I have access to a cluster I can run any additional experiments if needed.

vwxyzjn:
Hey, sorry folks for not replying sooner. I looked into the PR a bit more and looks like @jseppanen has already addressed my comments on 50M steps. Thanks a lot. I generated the following plots using https://github.com/vwxyzjn/ppo-atari-metrics/blob/main/rlops.py


```
python rlops.py --wandb-project-name envpool-atari \
    --wandb-entity openrlbenchmark \
    --filters 'ppo_dna_atari_envpool_94fc331?wpn=cleanrl&we=jseppanen' 'ppo_atari_envpool_xla_jax_truncation?metric=charts/avg_episodic_return'   \
    --env-ids BattleZone-v5 DoubleDunk-v5 NameThisGame-v5 Phoenix-v5 Qbert-v5 Pong-v5 BeamRider-v5 Breakout-v5 Tennis-v5 \
    --output-filename compare.png --scan-history
```

image

The results look good. **I will go ahead and run the Pong-v5 experiments to match the 10M steps, and that should be all for all of the experiments.** @jseppanen would you mind moving the runs from your entity to `openrlbenchmark`? You can move them as shown in the video below:

https://user-images.githubusercontent.com/5555347/202879526-c2b54798-a103-4192-a1f1-b3183abc176f.mov



# A note on environment preprocessing

The preprocessing steps should be like below, but `full_action_space=True` is not currently supported by EnvPool (https://github.com/sail-sg/envpool/issues/220). Let's put this note in the documentation and not block this PR any longer.

```python
envs = envpool.make(
    args.env_id,
    env_type="gym",
    num_envs=args.num_envs,
    episodic_life=False,  # Machado et al. 2017 (Revisitng ALE: Eval protocals) p. 6
    repeat_action_probability=0.25,  # Machado et al. 2017 (Revisitng ALE: Eval protocals) p. 12
    noop_max=1,  # Machado et al. 2017 (Revisitng ALE: Eval protocals) p. 12 (no-op is deprecated in favor of sticky action)
    max_episode_steps=int(108000 / 4),  # Hessel et al. 2018 (Rainbow DQN), Table 3, Max frames per episode
    reward_clip=True,
    seed=args.seed,
    # full_action_space=True, # currently not supported by EnvPool Machado et al. 2017 (Revisitng ALE: Eval protocals) Table 5
)
```

cleanrl/ppo_dna_atari_envpool.py
```diff
@@ -0,0 +1,426 @@
...
+        # Value network optimization
+        agent_value.train()
+        for epoch in range(args.value_update_epochs):
+            np.random.shuffle(b_inds)
+            for start in range(0, args.batch_size, args.value_batch_size):
+                end = start + args.value_batch_size
+                mb_inds = b_inds[start:end]
+
+                newvalue = agent_value.get_value(b_obs[mb_inds])
+
```
vwxyzjn, line 368: I hit a major regression with the latest script. It appears it's because `newvalue` here is not flattened

image

PR #264, opened 2022-08-22
chore: remove unused parameters in jax implementations

kinalmehta:
## Description

This PR refactors the *_jax.py implementations as follows:
- removes unused parameters from argparse (`torch-deterministic` as jax is deterministic by default, `cuda` as GPU is activated by default if available)
- fix pybullet envs in td3_jax and ddpg_jax

Given this is a **non-performance-impacting refactor**, re-running the benchmark is not required.


vwxyzjn (approved):

vwxyzjn:
Thank you! @kinalmehta

Issue #266, opened 2022-08-26
SAC discrete


timoklein:
Hey there!

I've used this repo's SAC code as starting point for an implementation of SAC-discrete ([paper](https://arxiv.org/pdf/1910.07207.pdf)) for a project of mine. If you're interested, I'd be willing to contribute it to cleanRL.

The differences to SAC for continuous action spaces aren't too big and I can start from a working implementation, so this shouldn't take too long.

What do you think?

## Checklist
- [x] I have installed dependencies via `poetry install` (see [CleanRL's installation guideline](https://docs.cleanrl.dev/get-started/installation/).
- [x] I have checked that there is no similar [issue](https://github.com/vwxyzjn/cleanrl/issues) in the repo (required)

vwxyzjn:
Hi @timoklein, thanks for being interested in submitting a contribution! SAC discrete indeed sounds like an interesting addition to CleanRL. I just glanced at the paper and would recommend prototyping a `sac_atari.py` to work with Atari games as done in the paper.  

I was a bit surprised to see the algorithm performs poorly on Pong. Do you have any insight on this? Maybe this is some implementation details stuff... CC @dosssman

timoklein:
> I just glanced at the paper and would recommend prototyping a sac_atari.py to work with Atari games as done in the paper.

Getting to work on it!

> I was a bit surprised to see the algorithm performs poorly on Pong. Do you have any insight on this? Maybe this is some implementation details stuff... CC @dosssman, who was the main contributor for the CleanRL's SAC implementation.

For reference, here are the reported results in the paper:
![SAC_discrete_results](https://user-images.githubusercontent.com/36668911/187127192-3583e752-4a4b-4159-81f6-63f3385427f7.png)

In my opinion, the bad results on Pong are due to the evaluation scheme. Evaluation at 100k time steps on Atari is in my opinion a very tough setting for a "standard" model-free RL (some newer methods like CURL or DrQ may perform better). Rainbow also doesn't improve over a random agent in this setting. Therefore we should focus the evaluation on games where meaningful improvements over a random baseline can be made, e.g. Seaquest, James Bond or Road Runner.

PR #270, opened 2022-08-29
SAC-discrete implementation

timoklein:
## Description
Adds the SAC-discrete algorithm as discussed in https://github.com/vwxyzjn/cleanrl/issues/266. 

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [x] New algorithm
- [x] Documentation

## Checklist:

- [x] I have contacted [vwxyzjn](https://github.com/vwxyzjn) to obtain access to the [openrlbenchmark W&B team](https://wandb.ai/openrlbenchmark) (**required**).
    - [x] I have added links to the original paper and related papers (if applicable).
    - [x] I have updated the overview sections at the [docs](https://docs.cleanrl.dev/rl-algorithms/overview/) and the [repo](https://github.com/vwxyzjn/cleanrl#overview)


cleanrl/sac_atari.py
```diff
@@ -0,0 +1,374 @@
...
+
+    def get_action(self, x):
+        logits = self(x / 255.0)
+        policy_dist = Categorical(logits=logits)
+        action = policy_dist.sample()
+        # Action probabilities for calculating the adapted soft-Q loss
+        action_probs = policy_dist.probs
+        # From https://github.com/p-christ/Deep-Reinforcement-Learning-Algorithms-with-PyTorch/blob/135d3e2e06bbde2868047d738e3fc2d73fd8cc93/agents/actor_critic_agents/SAC_Discrete.py#L58
+        # Add small numerical constant where action probability is 0 to avoid log(0)
+        z = action_probs == 0.0
```
Howuhh, line 195: From my experience, you can achieve the same with `F.log_softmax(logits, dim=-1)`, which is more explicit and convenient

timoklein: Thanks for the hint! Changed it.

cleanrl/sac_atari.py
```diff
@@ -0,0 +1,359 @@
...
+
+        return logits
+
+    def get_action(self, x):
+        logits = self(x / 255.0)
+        policy_dist = Categorical(logits=logits)
+        action = policy_dist.sample()
+        # Action probabilities for calculating the adapted soft-Q loss
+        action_probs = policy_dist.probs
+        log_pi = F.log_softmax(action_probs, dim=1)
```
Howuhh, line 181: small bug, should be logits, not action_probs (as probs are already after softmax)

cleanrl/sac_atari.py
```diff
@@ -0,0 +1,359 @@
...
+        return action, log_pi, action_probs
```
dosssman, line 182: Could we use `log_prob` instead, so that the naming are standardized across continuous and discrete action space tasks ?
https://github.com/vwxyzjn/cleanrl/blob/e466f6efb251462de2b80f064133f32cb5d83e22/cleanrl/sac_continuous_action.py#L143
Thanks.

cleanrl/sac_atari.py
```diff
@@ -0,0 +1,359 @@
...
+        env.seed(seed)
+        env.action_space.seed(seed)
+        env.observation_space.seed(seed)
+        return env
+
+    return thunk
+
+
+# He initialization
+def network_init(net, bias_const=0.0):
```
dosssman, line 104: Just nitpicking, but could we use `layer_init` instead, as in the other scripts:
https://github.com/vwxyzjn/cleanrl/blob/e466f6efb251462de2b80f064133f32cb5d83e22/cleanrl/ppo_continuous_action.py#L102

Thanks.

cleanrl/sac_atari.py
```diff
@@ -0,0 +1,359 @@
...
+    for param in net.parameters():
+        if type(param) == nn.Linear or type(param) == nn.Conv2d:
```
dosssman, line 106: I thought I would use this opportunity to learn a bit more, so let me know what you think:

While `type()` does the job, but from an OOP perspective, I feel like using `isinstance(param, nn.Linear)`
would be more adequate.

"type" usually refers to the "data type", life float, int, etc...
But `isinstance()` is expected when working with classes, such as `nn.Linear`.
Although in Python, they essentially work in the same way...

cleanrl/sac_atari.py
```diff
@@ -0,0 +1,359 @@
...
+    qf1_target = SoftQNetwork(envs).to(device)
+    qf2_target = SoftQNetwork(envs).to(device)
+    qf1_target.load_state_dict(qf1.state_dict())
+    qf2_target.load_state_dict(qf2.state_dict())
+    q_optimizer = optim.Adam(list(qf1.parameters()) + list(qf2.parameters()), lr=args.q_lr, eps=1e-4)
+    actor_optimizer = optim.Adam(list(actor.parameters()), lr=args.policy_lr, eps=1e-4)
+
+    # Automatic entropy tuning
+    if args.autotune:
+        target_entropy = -0.98 * torch.log(torch.tensor(envs.single_action_space.n, dtype=float))
```
dosssman, line 236: What's the reference for this implementation ?

dosssman:
Is there any benchmark run for this ? How does it perform ?

cleanrl/sac_atari.py
Howuhh: Yeah, I think max entropy should be -log(1 / |A|) (from entropy definition for discrete random variables). For example, that's what I have in my [implementation](https://github.com/Howuhh/hse_reinforcement_learning/blob/main/extra_algo/sac_discrete.ipynb) and [others](https://github.com/ku2482/sac-discrete.pytorch/blob/master/sacd/agent/sacd.py#L47)

cleanrl/sac_atari.py
timoklein: Indeed, thanks for the catch!

cleanrl/sac_atari.py
timoklein: You're both correct. I messed this up when going from numpy to torch

cleanrl/sac_atari.py
timoklein: From my knowledge I would agree with you that `isinstance()` is more Pythonic. As far as I know it also properly handles inheritance.
I started from the original code which was using `type()` and hadn't changed it. I'll clean this up.

cleanrl/sac_atari.py
timoklein: I'm definitely going to change to a `layer_init`-style function.
The paper does explicitly mention kaiming initialization, so I'll leave that. Is that okay with you?

timoklein:
@dosssman 

The results of the original paper are in https://github.com/vwxyzjn/cleanrl/issues/266. There's quite a number of games where its performance at 100k doesn't differ much from a random agent, e.g. Frostbite. I don't think it makes much sense evaluating those.

Right now I just ran it a couple of times manually to see that the code actually works. A modified version of this codebase has been able to solve [Catcher](https://pygame-learning-environment.readthedocs.io/en/latest/user/games/catcher.html) (PyGame) and some simple Minigrid environments.

In general, I'm going to try and find some good hyperparameters and then running it on a few environments where performance actually differs from a random agent. Don't know for sure when I have time for that though... Once that's done, I'll put a report in here. Maybe I'm going to run it for 200k steps also to verify the results.

After that plan on starting with some Docs :)

timoklein:
Posting an update:
I'm running 1m step experiments on Seaquest (takes a while) currently with two versions of the algorithm: One with an implementation as close as possible to cleanrl's SAC implementation and the other which is closer to the paper.  
The main difference is the update frequency: CleanRL does a critic update for every learning step, delays the actor updates but compensates for delayed actor updates. As far as I understand SAC-discrete does a single actor and critic update every four steps, not compensating for the delay.

timoklein:
I ran some experiments on `MsPacman` and `Seaquest`.

[Here's](https://wandb.ai/timo_kk/SAC-discrete/reports/SAC-discrete--VmlldzoyNjY0MzY3?accessToken=csgyf6higme5b9mfye95nm0ux3fu4q5jguw0c00u974lupw9p7ujkyhnqhvs1uy7) a link to a report with some results. The entropy regularization coefficient $\alpha$ has a tendency to explode when training longer but I couldn't find other experiments at 1m steps to verify whether that's an issue only in my code. Maybe @Howuhh and @dosssman have an idea?

This implementation doesn't quite match the results of the paper which might be due to not using evaluation mode (i.e. deterministic policy). If it's desired I can implement a test loop evaluating a deterministic policy.  

The performance does match the reported values in [this](https://github.com/toshikwa/sac-discrete.pytorch/issues/22) other implementation I found. 

Are there any other experiments you'd like me to run, e.g. specific environments or more seeds?

> I have tracked applicable experiments in [openrlbenchmark/cleanrl](https://wandb.ai/openrlbenchmark/cleanrl) with --capture-video flag toggled on (required).

Are there ways I can do this without `xvfb`? It's not on our group's machines and I don't have sudo.

If everything's fine, I'm gonna start writing Docs :)

vwxyzjn:
Hi @timoklein, thank you! The experiments look very interesting. 

> This implementation doesn't quite match the results of the paper which might be due to not using evaluation mode (i.e. deterministic policy). If it's desired I can implement a test loop evaluating a deterministic policy.

Unless it makes a huge difference in the reported results, I wouldn't worry about it.

> Are there any other experiments you'd like me to run, e.g. specific environments or more seeds?

Thanks for running the environments for three random seeds, which is great. For my personal interest, I would like to see the results in Pong, Breakout, and BeamRider, which are the common three environments that we used to benchmark other algorithms (see example [here](https://docs.cleanrl.dev/rl-algorithms/dqn/#experiment-results)).

>Are there ways I can do this without xvfb? It's not on our group's machines and I don't have sudo.

If this is an issue, feel free to run the experiments without xvfb. The video recordings are nice to have, but I am not requiring every run to have them. For example, it's not practical to obtain videos in EnvPool, so I didn't do it. 

That said, please save the model at the end of training so that we can visualize them later. Feel free to use the following code:

```python
torch.save(agent.state_dict(), f"models/{run_name}/agent.pt")
if args.prod_mode:
    wandb.save(f"models/{run_name}/agent.pt", base_path=f"models/{run_name}", policy="now")
```

Howuhh:
@timoklein Sometimes target entropy maybe just very high and hard to reach and the loss can explode (as alpha will grow and grow), so usually I tune a bit coefficient (which is 0.98 defualt in code right know)

timoklein:
> @timoklein Sometimes target entropy maybe just very high and hard to reach and the loss can explode (as alpha will grow and grow), so usually I tune a bit coefficient (which is 0.98 defualt in code right know)
> 
> Some discussion about it: [toshikwa/sac-discrete.pytorch#15 (comment)](https://github.com/toshikwa/sac-discrete.pytorch/issues/15#issuecomment-729547166)

Thank you very much for the link. The 0.98 is the author's default. I'll make a parameter out of that and tune it a little bit.

> Thanks for running the environments for three random seeds, which is great. For my personal interest, I would like to see the results in Pong, Breakout, and BeamRider, which are the common three environments that we used to benchmark other algorithms (see example [here](https://docs.cleanrl.dev/rl-algorithms/dqn/#experiment-results)).

Sure, running three environments is doable. Currently our server's a little busy but once that's done I can start the experiments.

> That said, please save the model at the end of training so that we can visualize them later. Feel free to use the following code:

Thanks, will do!

timoklein:
Tried a small set of entropy scaling factors (0.7, 0.8, 0.9, 0.98) and clearly the value of the original paper (0.98) seems rather suboptimal.
See the report [here](https://wandb.ai/timo_kk/SAC-discrete/reports/Effect-of-entropy-scale--VmlldzoyNzQ4MDY4?accessToken=0w4q1v5w4ytzfjmbjowmvdxwglf75tvi6ivgk7i02ypssyhh0vxiise0quoithq3).

Going forward with 0.8. After trying out the delayed update once more I'm going to do three seeds on Pong, Breakout, and BeamRider up to 4m steps.

cleanrl/sac_atari.py
```diff
@@ -0,0 +1,350 @@
...
+            layer_init(nn.Linear(output_dim, 512)),
+            nn.ReLU(),
+            layer_init(nn.Linear(512, envs.single_action_space.n)),
+        )
+
+    def forward(self, x):
+        x = self.conv(x)
+        logits = self.fc_logits(x)
+
+        return logits
```
vwxyzjn, line 172: Is this function used somewhere? If not, please remove it.

cleanrl/sac_atari.py
```diff
@@ -0,0 +1,350 @@
...
+        env = gym.wrappers.FrameStack(env, 4)
+        env.seed(seed)
+        env.action_space.seed(seed)
+        env.observation_space.seed(seed)
+        return env
+
+    return thunk
+
+
+# He initialization
```
vwxyzjn, line 104: Remove this line. The layer initialization should also be explained in the `implementation details` section of the docs. Where is this initialization scheme coming from?

cleanrl/sac_atari.py
```diff
@@ -0,0 +1,350 @@
...
+        if "FIRE" in env.unwrapped.get_action_meanings():
+            env = FireResetEnv(env)
+        env = ClipRewardEnv(env)
+        env = gym.wrappers.ResizeObservation(env, (84, 84))
+        env = gym.wrappers.GrayScaleObservation(env)
+        env = gym.wrappers.FrameStack(env, 4)
+        env.seed(seed)
+        env.action_space.seed(seed)
+        env.observation_space.seed(seed)
+        return env
```
vwxyzjn, line 99: The use of Atari wrappers should be noted in the implementation details section of the docs. Probably copy from https://cleanrl-93vza6xsa-vwxyzjn.vercel.app/rl-algorithms/ppo/#implementation-details_1

cleanrl/sac_atari.py
```diff
@@ -0,0 +1,350 @@
...
+
+    return thunk
+
+
+# He initialization
+def layer_init(layer, bias_const=0.0):
+    if isinstance(layer, nn.Linear) or isinstance(layer, nn.Conv2d):
```
vwxyzjn, line 106: The check here should not be necessary.

cleanrl/sac_atari.py
```diff
@@ -0,0 +1,350 @@
...
+    else:
+        alpha = args.alpha
+
+    rb = ReplayBuffer(
+        args.buffer_size,
+        envs.single_observation_space,
+        envs.single_action_space,
+        device,
+        optimize_memory_usage=True,
+        handle_timeout_termination=True,
```
vwxyzjn, line 243: Explain how truncation is handled like 

```
`sac_atari.py` handles truncation and termination properly like (Mnih et al., 2015)[^1] by using SB3's replay buffer's `handle_timeout_termination=True`.
```

See https://docs.cleanrl.dev/rl-algorithms/dqn/#implementation-details

cleanrl/sac_atari.py
```diff
@@ -0,0 +1,350 @@
...
+                writer.add_scalar("charts/episodic_return", info["episode"]["r"], global_step)
+                writer.add_scalar("charts/episodic_length", info["episode"]["l"], global_step)
+                break
+
+        # TRY NOT TO MODIFY: save data to reply buffer; handle `terminal_observation`
+        real_next_obs = next_obs.copy()
+        for idx, d in enumerate(dones):
+            if d:
+                real_next_obs[idx] = infos[idx]["terminal_observation"]
+        rb.add(obs, real_next_obs, actions, rewards, dones, [info])
```
vwxyzjn, line 273: Maybe do `rb.add(obs, real_next_obs, actions, rewards, dones, infos)` instead of `rb.add(obs, real_next_obs, actions, rewards, dones, [info])`?

vwxyzjn (commented):
Thank you @timoklein for this awesome contribution. The results and benchmark look very nice. 

One quick note is we typically put an implementation details section for each variant instead of merging them, as shown below.

image

timoklein:
Thank you very much for the detailed feedback. I'm going work everything in.

The contribution's going a bit slow because I also have other stuff to do but the process is very valuable to me. I have some other things in my mind that I want to contribute in the future so those will hopefully go a faster once I'm used to everything :)

cleanrl/sac_atari.py
```diff
@@ -0,0 +1,350 @@
...
+            with torch.no_grad():
+                _, next_state_log_pi, next_state_action_probs = actor.get_action(data.next_observations)
+                qf1_next_target = qf1_target(data.next_observations)
+                qf2_next_target = qf2_target(data.next_observations)
+                # we can use the action probabilities instead of MC sampling to estimate the expectation
+                min_qf_next_target = next_state_action_probs * (
+                    torch.min(qf1_next_target, qf2_next_target) - alpha * next_state_log_pi
+                )
+                # adapt Q-target for discrete Q-function
+                min_qf_next_target = min_qf_next_target.sum(dim=1)[..., None]
```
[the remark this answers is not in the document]
Howuhh: It's just a placeholder, so you don't have to enter all dims by hand. For example, if you have 4dim array:
1. select last: [..., idx] (not [:, :, :, idx])
2. select last 2: [..., idx1, idx2] (not [:, :, idx1, idx2])
3. add 1 dim at the end: [..., None] (not [:, :, :, :, None])

vwxyzjn:
@dosssman would you mind redoing the review for the PR given the recent changes to the algorithm and the documentation?

Chulabhaya:
Hi @timoklein ! This is awesome work! I was wondering, have you tested this on a non-Atari discrete environment? I was attempting to run this on discrete Lunar Lander and it was struggling to learn, so I wasn't sure if that was an error made on my involving hyperparameters or model architectures when going from the convolutional to an MLP architecture. Thanks!

vwxyzjn:
Executed some commands from  #307 

```
python -m cleanrl_utils.rlops_tags \
    --add pr-270 \
    --source-tag latest \
    --filters 'sac_atari' \
    --wandb-project-name cleanrl \
    --wandb-entity openrlbenchmark
python -m cleanrl_utils.rlops_tags \
    --remove rlops-pilot \
    --source-tag pr-270 \
    --filters 'sac_atari' \
    --wandb-project-name cleanrl \
    --wandb-entity openrlbenchmark
python -m cleanrl_utils.rlops_tags \
    --add v1.0.0b1-42-g4915e4c \
    --source-tag latest \
    --filters 'sac_atari' \
    --wandb-project-name cleanrl \
    --wandb-entity openrlbenchmark
python -m cleanrl_utils.rlops_tags \
    --add v1.0.0b1-43-g6f7251f \
    --source-tag latest \
    --filters 'sac_atari' \
    --wandb-project-name cleanrl \
    --wandb-entity openrlbenchmark
```

Then I did the following:

1. manually removes the `latest` tag from the runs that use 0.8 as entropy coefficient
1. remove the `v1.0.0b1-43-g6f7251f` tag from the runs that use 0.8 as entropy coefficient
1. remove the `v1.0.0b1-42-g4915e4c` tag from the runs that use 0.88 as entropy coefficient

As a result, the 

python -m cleanrl_utils.rlops --wandb-project-name cleanrl \
    --wandb-entity openrlbenchmark \
    --filters 'sac_atari?tag=pr-270&tag=v1.0.0b1-43-g6f7251f' \
    --env-ids BreakoutNoFrameskip-v4 PongNoFrameskip-v4 BeamRiderNoFrameskip-v4 \
    --output-filename compare.png --scan-history --report

Clone the report to the `openrlbenchmark/openrlbenchmark` name space and manually override the runset names to be `CleanRL's sac_atari.py`.

https://wandb.ai/openrlbenchmark/openrlbenchmark/reports/Atari-CleanRL-s-sac_atari-py--VmlldzoyOTYzNjg3

timoklein:
> Hi @timoklein ! This is awesome work! I was wondering, have you tested this on a non-Atari discrete environment? I was attempting to run this on discrete Lunar Lander and it was struggling to learn, so I wasn't sure if that was an error made on my involving hyperparameters or model architectures when going from the convolutional to an MLP architecture. Thanks!

Hey there! I've personally used it on two other environments: [Minigrid](https://minigrid.farama.org/) and [Catcher](https://pygame-learning-environment.readthedocs.io/en/latest/user/games/catcher.html).

On minigrid, I'm using the [RIDE](https://github.com/facebookresearch/impact-driven-exploration/blob/main/src/models.py)-encoder with `target_network_frequency: 20000`,  `target_entropy_scale: 0.8` (for an action space where useless actions are removed) and `update_frequency: 2`.

I haven't used it in non-pixel environments yet. In general I found the algorithm to be sensitive to the `target_entropy_scale`. If you are enforcing a more uniform policy (i.e. high entropy scale) but the action space contains actions that are rarely needed or even never needed, then it won't properly learn.

I'll try a little bit as well.

cleanrl/sac_atari.py
```diff
@@ -0,0 +1,339 @@
...
+
+
+def layer_init(layer, bias_const=0.0):
+    nn.init.kaiming_normal_(layer.weight)
+    torch.nn.init.constant_(layer.bias, bias_const)
+    return layer
+
+
+# ALGO LOGIC: initialize agent here:
+# NOTE: Sharing a CNN encoder between Actor and Critics is not recommended for SAC
```
dosssman, line 111: Don't they actually share the CNN encoder between critics and actor, but they stop the gradient going into the encoder from the actor's `fc1` and `fc_logits` though ?

Here is where they actually share the CNN between the Actor and Critics, if I am not mistaken.
https://github.com/denisyarats/pytorch_sac_ae/blob/7fa560e21c026c04bb8dcd72959ecf4e3424476c/sac_ae.py#L247

And here is where they detach the gradient of the actor layers so that it does not go into the shared CNN encoder:
https://github.com/denisyarats/pytorch_sac_ae/blob/7fa560e21c026c04bb8dcd72959ecf4e3424476c/sac_ae.py#L162

So I would say that the problem isn't really sharing the CNN encoder, but letting actor gradient backpropagate into the latter, as mentioned in the NOTE 2 lines below.

Since it seems to work in this implementation anyway, it should be fine to move ahead and merge.
It could be interesting to see how it fares with shared CNN encoder but stop_gradient in the actor level though.

dosssman: From the paper, Appendix B.2 for reference:
> The actor and critic networks both have separate encoders, although we share the weights of the conv layers between them. Furthermore, only the critic optimizer is allowed to update these weights (e.g. we truncate the gradients from the actor before they propagate to the shared conv layers).

dosssman (approved):
Looking good for me.
Good job getting it working.

araffin:
tagging myself @araffin for review (later this week probably)

cleanrl/sac_atari.py
```diff
@@ -0,0 +1,342 @@
...
+    qf1_target = SoftQNetwork(envs).to(device)
+    qf2_target = SoftQNetwork(envs).to(device)
+    qf1_target.load_state_dict(qf1.state_dict())
+    qf2_target.load_state_dict(qf2.state_dict())
+    q_optimizer = optim.Adam(list(qf1.parameters()) + list(qf2.parameters()), lr=args.q_lr)
+    actor_optimizer = optim.Adam(list(actor.parameters()), lr=args.policy_lr)
+
+    # Automatic entropy tuning
+    if args.autotune:
+        target_entropy = -args.target_entropy_scale * torch.log(1 / torch.tensor(envs.single_action_space.n))
```
araffin, line 221: where does that come from?
(both the entropy scale and the heuristic)

cleanrl/sac_atari.py
```diff
@@ -0,0 +1,342 @@
...
+        if global_step > args.learning_starts:
+            if global_step % args.update_frequency == 0:
+                data = rb.sample(args.batch_size)
+                # CRITIC training
+                with torch.no_grad():
+                    _, next_state_log_pi, next_state_action_probs = actor.get_action(data.next_observations)
+                    qf1_next_target = qf1_target(data.next_observations)
+                    qf2_next_target = qf2_target(data.next_observations)
+                    # we can use the action probabilities instead of MC sampling to estimate the expectation
+                    min_qf_next_target = next_state_action_probs * (
```
araffin, line 278: do you have a reference for that too? (although it does make sense)

dosssman: Straight from the paper: https://arxiv.org/pdf/1910.07207.pdf

image

cleanrl/sac_atari.py
dosssman: Looks like the definition in the Discrete SAC paper: https://arxiv.org/pdf/1910.07207.pdf

image

image

Howuhh: This is common wisdom already, I guess. You can see discussion about scale here https://github.com/toshikwa/sac-discrete.pytorch/issues/15#issuecomment-729547166, heuristic is just max entropy for categorical distribution

cleanrl/sac_atari.py
```diff
@@ -0,0 +1,342 @@
...
+                qf2_loss = F.mse_loss(qf2_a_values, next_q_value)
+                qf_loss = qf1_loss + qf2_loss
+
+                q_optimizer.zero_grad()
+                qf_loss.backward()
+                q_optimizer.step()
+
+                # ACTOR training
+                _, log_pi, action_probs = actor.get_action(data.observations)
+                qf1_pi = qf1(data.observations)
```
araffin, line 300: it's a bit weird to call it `qf1_pi` here as it doesn't depends on the actor.

To make it even more explicit, I guess you could even do a `min_qf_pi.detach()`.

cleanrl/sac_atari.py
Howuhh: That's how the original paper about SAC-discrete does it

dosssman:
@timoklein Here is a snippet that would address @araffin's comments:
https://github.com/vwxyzjn/cleanrl/pull/270#discussion_r1031332675
> why do you keep dim here as you flatten it in the next line?

https://github.com/vwxyzjn/cleanrl/pull/270#discussion_r1031337608
> it's a bit weird to call it `qf1_pi` here as it doesn't depends on the actor.
> 
> To make it even more explicit, I guess you could even do a `min_qf_pi.detach()`.

https://github.com/vwxyzjn/cleanrl/pull/270#discussion_r1031338611
> you could re-use `log_pi, action_probs` from before, no? (don't forget the `detach()` in that case)

It essentially reduces the amount of calls to the Q network and the policy (especially if using entropy autotuning) by re-using the first call and leveraging `.detach()` to stop gradients where necessary..
This will also require to use a joint loss `total_loss = actor_loss + qf_loss` so that `.backward()` is called only once (otherwise it will probably throw an `gradient was at version XX but ...` error.
However, it might make the code a bit harder to understand for beginners though, so not sure if this is preferable to the current method. Any thoughts @vwxyzjn regarding the complexity of this factored variant ?

```python
if global_step > args.learning_starts and if global_step % args.update_frequency == 0:
    data = rb.sample(args.batch_size)
    # CRITIC training
    with torch.no_grad():
        _, next_state_log_pi, next_state_action_probs = actor.get_action(data.next_observations)
        qf1_next_target = qf1_target(data.next_observations)
        qf2_next_target = qf2_target(data.next_observations)
        # we can use the action probabilities instead of MC sampling to estimate the expectation
        min_qf_next_target = next_state_action_probs * (
            torch.min(qf1_next_target, qf2_next_target) - alpha * next_state_log_pi
        )
        # adapt Q-target for discrete Q-function
        min_qf_next_target = min_qf_next_target.sum(dim=1)
        next_q_value = data.rewards + (1 - data.dones) * args.gamma * (
            min_qf_next_target
        ) # NOTE: double check the output dimension

    # use Q-values only for the taken actions
    qf1_values = qf1(data.observations)
    qf2_values = qf2(data.observations)
    qf1_a_values = qf1_values.gather(1, data.actions.long()).view(-1)
    qf2_a_values = qf2_values.gather(1, data.actions.long()).view(-1)
    qf1_loss = F.mse_loss(qf1_a_values, next_q_value)
    qf2_loss = F.mse_loss(qf2_a_values, next_q_value)
    qf_loss = qf1_loss + qf2_loss

    # ACTOR training
    _, log_pi, action_probs = actor.get_action(data.observations)
    min_qf_values = torch.min(qf1_values, qf2_values)
    # no need for reparameterization, the expectation can be calculated for discrete actions
    # NOTE regarding joint loss optim: detaching qf values.
    actor_loss = (action_probs * ((alpha * log_pi) - min_qf_values.detach())).mean()

    # Loss for joint actor and network training
    total_loss = qf_loss + actor_loss

    if args.autotune:
        # use action probabilities for temperature loss
        # NOTE regarding joint loss optim: detaching log pi fro the policy
        alpha_loss = (action_probs * (-log_alpha * (log_pi .detach()+ target_entropy))).mean()

        total_loss += alpha_loss

        a_optimizer.zero_grad()

    # NOTE: call .backward() only once to compute gradients for all the components
    q_optimizer.zero_grad()
    actor_optimizer.zero_grad()
    total_loss.backward()
    q_optimizer.step()
    actor_optimizer.step()
    
    if args.autotune:
        a_optimizer.step()
        alpha = log_alpha.exp().item()
```

Hope it helps.

araffin:
> This will also require to use a joint loss total_loss = actor_loss + qf_loss so that .backward() is called only once 

you could detach `action_probs` too and no joint loss is needed then?

dosssman:
Indeed.

Unlike the sac continuous, the output of the policy ia not fed through the Q functions for the actor loss, so joint loss optimization will not be required.
________________________________

timoklein:
Thanks for the input everyone!

I'm going to produce a version based on @dosssman 's snippet.
```python
min_qf_next_target = min_qf_next_target.sum(dim=1)
        next_q_value = data.rewards + (1 - data.dones) * args.gamma * (
            min_qf_next_target
        ) # NOTE: double check the output dimension
```

You either have to flatten or keep the dim in the sum to not do an outer product. I'll do flatten because it minimizes the diff between this and `sac_continuous`. Same reason is why I'm going to not opt for the joint loss.

Since there's been a bunch of changes now (e.g. diff minimization, https://github.com/vwxyzjn/cleanrl/pull/323 and now these) I'll also re-run the experiments to make sure that we have proper results. That's gonna take over the weekend at least.

timoklein:
The codespell check complains about the name "Jimmy Ba". It doesn't seem to have inline ignores ([issue](https://github.com/codespell-project/codespell/issues/1212)), so I'm not quite sure how to handle it.

TODOS:

- [ ] Change doc links to main branch.
- [ ] Update plots with new results.
- [ ] Update report with new results.

dosssman:
> eps=1e-4 for Adam is required. Without this, there are seeds where SAC-d doesn't learn at all on Pong. This setting is also used in the [author's codebase](https://github.com/p-christ/Deep-Reinforcement-Learning-Algorithms-with-PyTorch/blob/master/agents/actor_critic_agents/SAC.py). Has cost me a while to figure this one out...

Nice catch. That is quite an innocuous implementation detail which actually has a drastic impact.

> - Re-using log_pi and action_probs for the alpha loss will result in slower learning (but increase throughput considerably). I would therefore like to keep the code as it its. This also keeps it more close to the continuous SAC implementation.

Fair enough. We can review this if we go over SAC algorithm in general at a later point.

> - One could use no_grad/detach for the Q-values used in the actor loss. This yields a 5% speedup in terms of SPS with minuscule effects on learning on Pong. The original codebase doesn't use no_grad (see [here](https://github.com/p-christ/Deep-Reinforcement-Learning-Algorithms-with-PyTorch/blob/135d3e2e06bbde2868047d738e3fc2d73fd8cc93/agents/actor_critic_agents/SAC.py#L172)). I don't think 5% is worth deviating from the reference implementation. Re-using Q-values from the critic loss is a bad idea and hurts learning speed a lot.

Interesting. Could you please share the code for the variant that re-uses Q networks ?

timoklein:
> Nice catch. That is quite an innocuous implementation detail which actually has a drastic impact.

Indeed. I know some papers who go about tuning it, most don't. Rainbow also has a non-standard setting afaik (`1.5e-4`). Found a discussion [here](https://arxiv.org/pdf/1910.05446.pdf), though it's not particularly aimed at RL.

> Interesting. Could you please share the code for the variant that re-uses Q networks ?

Looks like this. Fetching new values with `no_grad` is not an issue.

```python
# CRITIC training
with torch.no_grad():
    _, next_state_log_pi, next_state_action_probs = actor.get_action(data.next_observations)
    qf1_next_target = qf1_target(data.next_observations)
    qf2_next_target = qf2_target(data.next_observations)
    # we can use the action probabilities instead of MC sampling to estimate the expectation
    min_qf_next_target = next_state_action_probs * (
        torch.min(qf1_next_target, qf2_next_target) - alpha * next_state_log_pi
    )
    # adapt Q-target for discrete Q-function
    min_qf_next_target = min_qf_next_target.sum(dim=1)
    next_q_value = data.rewards.flatten() + (1 - data.dones.flatten()) * args.gamma * (min_qf_next_target)

# use Q-values only for the taken actions
qf1_values = qf1(data.observations)
qf2_values = qf2(data.observations)
qf1_a_values = qf1_values.gather(1, data.actions.long()).view(-1)
qf2_a_values = qf2_values.gather(1, data.actions.long()).view(-1)
qf1_loss = F.mse_loss(qf1_a_values, next_q_value)
qf2_loss = F.mse_loss(qf2_a_values, next_q_value)
qf_loss = qf1_loss + qf2_loss

q_optimizer.zero_grad()
qf_loss.backward()
q_optimizer.step()

# ACTOR training
_, log_pi, action_probs = actor.get_action(data.observations)
# NO NEW VALUES HERE!
# qf1_values = qf1(data.observations)
# qf2_values = qf2(data.observations)
min_qf_values = torch.min(qf1_values, qf2_values).detach()
# no need for reparameterization, the expectation can be calculated for discrete actions
actor_loss = (action_probs * ((alpha * log_pi) - min_qf_values)).mean()
```

It's a bit hand-wavy, but my best explanation is that not using up-to-date values will result in steps that are slightly off. Over time, these errors accumulate and throw off training. I would assume the effect is more pronounced in later stages on Pong due to the environment structure: No or slow progress at the start, then the agent figures out the environment and rapidly solves it.

dosssman:
> It's a bit hand-wavy, but my best explanation is that not using up-to-date values will result in steps that are slightly off. Over time, these errors accumulate and throw off training. I would assume the effect is more pronounced in later stages on Pong due to the environment structure: No or slow progress at the start, then the agent figures out the environment and rapidly solves it.

Well, that might be one reason. Although intuitively, it does not seem like just "one Q network" update would affect the critic score used to update the policy. The difference in outputs of Qf1 and QF2 before and after one update step are probably marginal.

In any case, I concur we can probably leave the default implementation to stay consistent with the reference implementations, as well as the SAC continuous variant.
Later, it might be interesting to investigate that aspect a bit more, albeit on a simpler task without pixel-observations first, I think.

Great work. Overall looking good to me.

timoklein:
@dosssman 

Monday's results left me no peace because they were just run on a single seed. So I did actually run some more experiments considering the different types of update rules. (Not) Surprisingly, the results changed:
- **all**: Always get most up to date values for all loses. Backprop through Q-values in the actor loss (like [here](https://github.com/p-christ/Deep-Reinforcement-Learning-Algorithms-with-PyTorch/blob/135d3e2e06bbde2868047d738e3fc2d73fd8cc93/agents/actor_critic_agents/SAC.py#L172)). This is also the slowest version at ~77 long-run SPS.
- **nograd_Q**: Get new Q-values for the actor loss by calling the critics, but don't backprop through them. This gives a big boost in throughput, running at around 94 long-run SPS.
- **reuse_Q**: Re-use the Q-values used in the critic loss for the actor loss with `.detach()` (see the post above). Is even faster at ~96 long-run SPS. Problem here is that it destabilizes training: One seed doesn't learn at all.
- **nograd_Q_reuse_alpha**: Get new Q-values for the actor without backprop. Also re-uses the log-probabilities and action probabilities from the actor loss with `.detach()` in the alpha loss. Fastest version in terms of SPS at ~99.

sac_d_update_ablations

Right now I'm running 5M experiments for the openrlbenchmark with `nograd_Q_reuse_alpha`. Let's see what happens on the other games since I only looked at Pong. It's probably also game-dependent but since this version is also the fastest one, I'll keep it.

Thanks for all the input and ideas! Once the experiments are done, I'm going to update the results and remaining todos and we can get this wrapped up :)

dosssman:
@Chulabhaya 

Original SAC implementation was developed for continuous action case, while this one is for discrete action, hence the difference in computing target entropy. Furthermore, in both case target entropy is a "heuristic".

Maybe following down this thread can bring some clarity: https://github.com/vwxyzjn/cleanrl/pull/270#discussion_r1031334739

dosssman:
@timoklein 
Thanks a lot for the detailed experiments.
Indeed, running experiment with more than one seeds is critical. Sometimes, some seed can lead to "degenerate" results that are not really related to the algorithm itself.

> reuse_Q: Re-use the Q-values used in the critic loss for the actor loss with .detach() (see the post above). Is even faster at ~96 long-run SPS. Problem here is that it destabilizes training: One seed doesn't learn at all.

From the plot, I notice that there is a large variance in the `reuse_Q` variant. I suspect there might be one or two seed that are messing up the overall result, so it might be worth checking the seeds one by one.

Overall, I think re-using at least log_pi for alpha optimization (if applicable) is a reasonable trade-off (should also be doable on the continuous variant).

timoklein:
@dosssman 

Yes, re-using log_pi and the action probs is what I'm doing now.

Indeed, most of high variance and bad performance of the `reuse_Q` method comes from a single seed (seed=1). Removing it makes the method competitive, but arguably still worse to the version I'm currently using. As it's the only update variant failing to learn on this seed, we might also find stability issues on other seeds that I haven't evaluated.

cleanRL_sacd_removed_seed

dosssman:
Great contribution !
I think we can proceed to merging this as it matches baseline results while maintaining a relatively simple implementation.
Other issues, if any exists, will come in light as other people try it out from a fresher perspective.

araffin (commented):
minor: the title for the pong figure is not the right one

timoklein:
> minor: the title for the pong figure is not the right one

Thanks, good catch!

timoklein:
Anything more to do here for me? If this is blocking

> The codespell check complains about the name "Jimmy Ba". It doesn't seem to have inline ignores (https://github.com/codespell-project/codespell/issues/1212), so I'm not quite sure how to handle it.

I can just remove the citation so that the CI error is gone.

cleanrl/sac_atari.py
```diff
@@ -0,0 +1,342 @@
...
+        help="the learning rate of the Q network network optimizer")
+    parser.add_argument("--update-frequency", type=int, default=4,
+        help="the frequency of training updates")
+    parser.add_argument("--target-network-frequency", type=int, default=8000,
+        help="the frequency of updates for the target networks")
+    parser.add_argument("--alpha", type=float, default=0.2,
+        help="Entropy regularization coefficient.")
+    parser.add_argument("--autotune", type=lambda x:bool(strtobool(x)), default=True, nargs="?", const=True,
+        help="automatic tuning of the entropy coefficient")
+    parser.add_argument("--target-entropy-scale", type=float, default=0.9,
```
vwxyzjn, line 73: The experiment results reported using `target-entropy-scale=0.89`, so maybe change it to 0.89 here to be consistent? 

image

vwxyzjn (approved):
Everything LGTM. Feel free to merge after you have resolved the minor `target-entropy-scale` issue. You should already have contributor access. Thanks so much for this contribution, and sorry for the delay.

> I can just remove the citation so that the CI error is gone.

Please keep the citation and just add ignore words in the pre-commit configs instead. https://github.com/vwxyzjn/cleanrl/blob/3f5535cab409a34e9f071c10b96a234925d8a8d5/.pre-commit-config.yaml#L38

timoklein:
> Everything LGTM. Feel free to merge after you have resolved the minor target-entropy-scale issue. You should already have contributor access. Thanks so much for this contribution, and sorry for the delay.

No problem. It's been a fun experience and I learned a lot. Looking forward to contributing more in the future!

timoklein:
@vwxyzjn 
I'm not quite sure why this still fails
https://github.com/vwxyzjn/cleanrl/actions/runs/3912194839/jobs/6686505080#step:4:92

"Ba" should be correctly added to the pre-commit
https://github.com/vwxyzjn/cleanrl/blob/c3fc57db7d599fc0e531c932f2cc916eceedea4a/.pre-commit-config.yaml#L38

I'm going to fix it but it might be Sunday before I get around to doing it.

PR #262, opened 2022-08-22
Added TQC

AdityaGudimella:
## Description
Closes #258. Implement [Truncated Quantile Critics](https://paperswithcode.com/paper/controlling-overestimation-bias-with)

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [x] New algorithm
- [ ] Documentation

## Checklist:

- [x] I've read the [CONTRIBUTION](https://github.com/vwxyzjn/cleanrl/blob/master/CONTRIBUTING.md) guide (**required**).
- [x] I have ensured `pre-commit run --all-files` passes (**required**).
- [ ] I have updated the documentation and previewed the changes via `mkdocs serve`.
- [ ] I have updated the tests accordingly (if applicable).

If you are adding new algorithms or your change could result in performance difference, you may need to (re-)run tracked experiments. See https://github.com/vwxyzjn/cleanrl/pull/137 as an example PR. 
- [ ] I have contacted [vwxyzjn](https://github.com/vwxyzjn) to obtain access to the [openrlbenchmark W&B team](https://wandb.ai/openrlbenchmark) (**required**).
- [ ] I have tracked applicable experiments in [openrlbenchmark/cleanrl](https://wandb.ai/openrlbenchmark/cleanrl) with `--capture-video` flag toggled on (**required**).
- [ ] I have added additional documentation and previewed the changes via `mkdocs serve`.
    - [ ] I have explained note-worthy implementation details.
    - [ ] I have explained the logged metrics.
    - [ ] I have added links to the original paper and related papers (if applicable).
    - [ ] I have added links to the PR related to the algorithm.
    - [ ] I have created a table comparing my results against those from reputable sources (i.e., the original paper or other reference implementation).
    - [ ] I have added the learning curves (in PNG format with `width=500` and `height=300`).
    - [ ] I have added links to the tracked experiments.
    - [ ] I have updated the overview sections at the [docs](https://docs.cleanrl.dev/rl-algorithms/overview/) and the [repo](https://github.com/vwxyzjn/cleanrl#overview)
- [ ] I have updated the tests accordingly (if applicable).

Issue #258, opened 2022-08-13
Add TQC to CleanRL


AdityaGudimella:
## Problem Description
Would you be interested in adding Truncated Quantile Critics to CleanRL? If so I can work on a PR.

If you are interested, then I have a few questions:
1. Is it okay if I use the new gym.Env api where `env.step` returns `obs, reward, terminated, truncated, info` instead of `obs, reward, done, info`?
2. Is it okay if I deviate slightly from the style followed by the other algorithms in CleanRL? I would like to put the code for calculating the losses and updating the models into a separate class. Everything else will remain the same. This way, if people are familiar with the general RL procedure, and only want to see how the core of the algorithm changes, they will only need to look at this class.

## Checklist
- [x] I have installed dependencies via `poetry install` (see [CleanRL's installation guideline](https://docs.cleanrl.dev/get-started/installation/).
- [x] I have checked that there is no similar [issue](https://github.com/vwxyzjn/cleanrl/issues) in the repo (required)

## Current Behavior
Algorithm is not present

## Expected Behavior
Algorithm will be present

## Steps to Reproduce
1. Look for list of supported algorithms in CleanRL.
3. Find that TQC is not present.

vwxyzjn:
Hi, @AdityaGudimella thanks for your interest in contribution! We would definitely be interested in the TQC contribution.

> Is it okay if I use the new `gym.Env` api where `env.step` returns `obs, reward, terminated, truncated, info` instead of `obs, reward, done, info`?

It's ok but note the merging timeline may be considerably delayed because of this. Because `poetry` only supports using a single version for `gym`, we would need to make sure other scripts at least run with the latest gym. Judging by the list of things gym recently breaks, it's hard to estimate the amount of work involved... We might be able to merge it sooner with an ad-hoc instruction like `poetry run pip install gym==0.24.1` in the usage section as a temporary solution. See example [here](https://docs.cleanrl.dev/rl-algorithms/ppo/#usage_8)

> Is it okay if I deviate slightly from the style followed by the other algorithms in CleanRL? I would like to put the code for calculating the losses and updating the models into a separate class. Everything else will remain the same. This way, if people are familiar with the general RL procedure and only want to see how the core of the algorithm changes, they will only need to look at this class.

I still think it's worth it to adhere to the styles in other scripts for consistency unless the loss function is re-used multiple times. The problems with putting loss function in the class are 
1. the class will have much more lines of code
[...]

PR #272, opened 2022-09-16
Draft: DroQ and TD3+TQC jax implementation

araffin:
## Description

FYI: unpolished jax implementation of TD3+DroQ and TD3+TQC implementations.
Related to #262 #258 
My plan is to try to have sac in jax, but currently jax rely on tensorflow for probability distributions :/
So I adapted TD3 instead.
I also want to make it even faster but would need to tweak a bit the way the replay buffer is used.

Reference:
- jax: https://github.com/ikostrikov/walk_in_the_park
- original: https://github.com/TakuyaHiraoka/Dropout-Q-Functions-for-Doubly-Efficient-Reinforcement-Learning/tree/main/KUCodebase/code
- SB3: https://github.com/Stable-Baselines-Team/stable-baselines3-contrib/pull/100

Known difference with original implementation: qf are updated at the same time of the actor instead of after each gradient step.

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] New algorithm
- [ ] Documentation

## Checklist:

### DIFF
--- a/cleanrl/td3_droq_continuous_action_jax.py
+++ b/cleanrl/td3_droq_continuous_action_jax.py
@@ -1,0 +1,322 @@
+# docs and experiment results can be found at https://docs.cleanrl.dev/rl-algorithms/td3/#td3_continuous_action_jaxpy
+import argparse
+import os
+import random
+import time
+from distutils.util import strtobool
+from typing import Sequence
+
+import flax
+import flax.linen as nn
+import gym
+import jax
+import jax.numpy as jnp
+import numpy as np
+import optax
+import pybullet_envs  # noqa
+from flax.training.train_state import TrainState
+from stable_baselines3.common.buffers import ReplayBuffer
+from torch.utils.tensorboard import SummaryWriter
+
+
+def parse_args():
+    # fmt: off
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--exp-name", type=str, default=os.path.basename(__file__).rstrip(".py"),
+        help="the name of this experiment")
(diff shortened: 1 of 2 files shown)


vwxyzjn (commented):
👀 how does Adan perform?

araffin:
> eyes how does Adan perform?

Results are very preliminary, ADAN performs on par or slightly better than ADAM, but nothing significant yet.
The noticeable difference is the FPS though (adan slower, for instance 100 FPS vs 130 FPS).
Btw, I managed to JIT the for loop =) (it goes 2x faster now but results are different than without jit :eyes: )

vwxyzjn:
FYI https://github.com/deepmind/distrax might be a better replacement for tensorflow probability

joaogui1:
fwiw you can also use tensorflow_probability with a jax backend and then you don't need to use tensorflow at all (in one of their tutorials they even explicitly unninstall tf)

araffin:
@vwxyzjn Good news, I've got a TQC + SAC version working =) (currently doing some runs)

@joaogui1 thanks, I gave distrax a try but it was giving me weird errors, and at the end it still depends on tf proba (which doesn't require tensorflow as I learned =)), so I switched to tf proba ;)

araffin:
Fyi, I converted that single file to a proof of concept of SB3 + Jax (SBX): https://github.com/araffin/sbx
the nice thing is that I'm re-using SB3 base class, which means it has access to saving/loading/scikit interface/callbacks and soon the RL zoo =)

PR #271, opened 2022-09-10
Poetry 1.2

vwxyzjn:
## Description
Adopting poetry 1.2 support. The locking and adding dependencies are orders of magnitude faster!

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] New algorithm
- [ ] Documentation

## Checklist:

- [x] I've read the [CONTRIBUTION](https://github.com/vwxyzjn/cleanrl/blob/master/CONTRIBUTING.md) guide (**required**).
- [x] I have ensured `pre-commit run --all-files` passes (**required**).
- [x] I have updated the documentation and previewed the changes via `mkdocs serve`.
- [x] I have updated the tests accordingly (if applicable).

PR #277, opened 2022-09-25
Update to support Gymnasium

### LINKED ISSUE #263: Upgrade gym version to 0.26.1
## Problem Description
Upgrade gym version used in cleanrl from 0.23.1 to 0.25.1

## Checklist
- [ ] I have installed dependencies via `poetry install` (see [CleanRL's installation guideline](https://docs.cleanrl.dev/get-started/installation/).
- [x] I have checked that there is no similar [issue](https://github.com/vwxyzjn/cleanrl/issues) in the repo (required)

## Possible Solution
StabeBaselines' ReplayBuffer currently does not support the new format returned by `gym.Env.step`. Their step api changed from:
`obs, rew, done, info = env.step(action)` to `obs, rew, terminated, truncated, info = env.step(action)`.
We would need to implement a slightly modified version of the ReplayBuffer in cleanRL itself. Other than this, the changes required are minimal.

I can submit an initial PR with changes required for SAC if you're interested.

arjun-kg:
## Description
A draft PR updating CleanRL to support Gymnasium. Closes #263 

This mostly includes updating step and seed API. Tries to use gymnasium branches on the dependent packages (SB3 etc) After these are updated, will verify the changes, check the tests, and get the PR ready for review. 

## Costa's comment:
Thanks @arjun-kg for the PR. We look forward to supporting the next generation of gym. 

It's important to identify the **performance-impacting changes** and **non-performance-impacting changes**:

* **non-performance-impacting changes**  is fine and we can just merge them straightaway
* **performance-impacting changes** is more complicated — we would need to run benchmark experiments to ensure there is no performance regression. We need to do this because *even bug fixes could make the performance worse* (see https://github.com/openai/gym/pull/2762#issuecomment-1135362092 as an example). 

In this PR for initial support fo v0.26.1, let's aim to make only **non-performance-impacting changes**. With that said, here is a todo list:
- [x] Deprecate pybullet (since the new mujoco environments are being maintained again but pybullet is not)
- [ ] Temporarily remove Isaac Gym Support
- [ ] re-run mojoco experiments (use the v4 environments instead of the current v2 environments)
- [ ] Atari wrappers upstream fixes (https://github.com/DLR-RM/stable-baselines3/pull/780)
- [ ] procgen

## Checklist:


vwxyzjn:
Related to #271

cleanrl/ppo.py
```diff
@@ -213,18 +213,18 @@ def get_action_and_value(self, x, action=None):
                     writer.add_scalar("charts/episodic_length", item["episode"]["l"], global_step)
                     break
```
vwxyzjn, line 214: This part may need to be changed to 

```
            if "episode" in info:
                for item in info["episode"]["r"]:
                    print(f"global_step={global_step}, episodic_return={item}")
                    writer.add_scalar("charts/episodic_return", item, global_step)
                    break
                for item in info["episode"]["l"]:
                    writer.add_scalar("charts/episodic_length", item, global_step)
                    break
```

cleanrl/c51.py
```diff
@@ -159,12 +159,12 @@ def linear_schedule(start_e: float, end_e: float, duration: int, t: int):
         envs.single_observation_space,
         envs.single_action_space,
         device,
-        handle_timeout_termination=True,
+        handle_timeout_termination=False,
```
vwxyzjn, line 162: Perfect!

cleanrl/ppo.py
```diff
@@ -213,18 +213,18 @@ def get_action_and_value(self, x, action=None):
...
+        # bootstrap value if not terminated
         with torch.no_grad():
             next_value = agent.get_value(next_obs).reshape(1, -1)
             if args.gae:
                 advantages = torch.zeros_like(rewards).to(device)
                 lastgaelam = 0
                 for t in reversed(range(args.num_steps)):
                     if t == args.num_steps - 1:
-                        nextnonterminal = 1.0 - next_done
+                        nextnonterminal = 1.0 - next_terminated
```
vwxyzjn, line 224: A note for myself: this is a change that could impact performance. We would need to re-run the benchmark here.

vwxyzjn (commented):
Hey @arjun-kg thanks for preparing the PR. Looking forward to using the latest gym. I left some very preliminary comments. 

One important thing to note during the refactor is to see if the change could result in a performance difference (not just a simple variable renaming). For example, the current PPO scripts did not handle the time out correctly, so handling time out correctly in this PR is a **performance-impacting** change. 

We need to be careful with the **performance-impacting** changes because we would need to re-run the benchmarks on those changes to ensure there is no surprise regression in the performance.

cleanrl/ppo.py
DavidSlayback: To replicate the original behavior, you probably need something like (but hopefully better-looking than!):
```python
if "episode" in info:
    first_idx = info["_episode"].nonzero()[0][0]
    r = info["episode"]["r"][first_idx]
    l = info["episode"]["l"][first_idx]
    print(f"global_step={global_step}, episodic_return={r}")
    writer.add_scalar("charts/episodic_return", r, global_step)
    writer.add_scalar("charts/episodic_length", l, global_step)
```
There's no guarantee that the first index in "episode" won't just be a zero, need the mask to specify which one.

Alternatively, it might be better to track a running average using the deques built into the RecordEpisodeStatistics wrapper, though that would likely results in different performance graphs.

vwxyzjn:
Thanks @arjun-kg for the PR. We look forward to supporting the next generation of gym. 

It's important to identify the **performance-impacting changes** and **non-performance-impacting changes**:

* **non-performance-impacting changes**  is fine and we can just merge them straightaway
* **performance-impacting changes** is more complicated — we would need to run benchmark experiments to ensure there is no performance regression. We need to do this because *even bug fixes could make the performance worse* (see https://github.com/openai/gym/pull/2762#issuecomment-1135362092 as an example). 

In this PR for initial support for v0.26.1, let's aim to make only **non-performance-impacting changes**. With that said, I have added a todo list in the PR description.

vwxyzjn:
@arjun-kg I made the first pass of editing to make `ppo.py` and `dqn.py` to pass CI. Could you try looking at `ddpg_continuous_action.py`, TD3, and DDPG?

Btw the plan is to have an announcement like the following on the main page, since I expect to encounter more issues. 

image

GaetanLepage:
Hi !
When reading through the proposed changes, I am not sure to understand the following:
Why do you replace `done` with `terminated` and not `terminated or truncated` ?
I am not sure to get why the `truncated` return value is ignored.

vwxyzjn:
@arjun-kg we are thinking of supporting both gymnasium and gym at the same time. See https://github.com/vwxyzjn/cleanrl/pull/318#issuecomment-1314666112 as an example.

arjun-kg:
@vwxyzjn sounds good, will check it out. I'm a bit tied up this week. I'll continue work on this from next week if it's okay.

cleanrl/sac_continuous_action.py
```diff
@@ -232,12 +231,13 @@ def get_action(self, x):
                 writer.add_scalar("charts/episodic_length", info["episode"]["l"], global_step)
                 break
 
-        # TRY NOT TO MODIFY: save data to reply buffer; handle `terminal_observation`
-        real_next_obs = next_obs.copy()
-        for idx, d in enumerate(dones):
-            if d:
-                real_next_obs[idx] = infos[idx]["terminal_observation"]
-        rb.add(obs, real_next_obs, actions, rewards, dones, infos)
```
ffelten, line 240: Hi, I guess this line has been forgotten in the migration :-).

pseudo-rnd-thoughts:
We have just released Gymnasium v0.27.0, this should be backward compatible. Would it be possible to update this Pr to v0.27 and check that nothing new breaks

arjun-kg:
@vwxyzjn recently SB3 supports gymnasium with a branch, but I'm not sure if some parallel work is going on to update cleanrl to gymnasium? Would you like me to update this PR to gymnasium with SB3 on the gymnasium branch?

cleanrl/ddpg_continuous_action.py
```diff
@@ -71,7 +70,7 @@ def thunk():
         if capture_video:
             if idx == 0:
                 env = gym.wrappers.RecordVideo(env, f"videos/{run_name}")
-        env.seed(seed)
+
         env.action_space.seed(seed)
         env.observation_space.seed(seed)
```
valentin-cnt, line 75: I think `env.observation_space.seed(seed)` can be remove

cleanrl/ddpg_continuous_action.py
```diff
@@ -191,12 +190,12 @@ def forward(self, x):
                 writer.add_scalar("charts/episodic_length", info["episode"]["l"], global_step)
                 break
 
-        # TRY NOT TO MODIFY: save data to reply buffer; handle `terminal_observation`
+        # TRY NOT TO MODIFY: save data to reply buffer; handle `final_observation`
         real_next_obs = next_obs.copy()
-        for idx, d in enumerate(dones):
+        for idx, d in enumerate(terminateds):
```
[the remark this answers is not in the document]
arjun-kg: Yes, this was a mistake, it should be `truncated`

kvrban:
Just tried the PR with: 

```diff
diff --git a/cleanrl/dqn.py b/cleanrl/dqn.py
 import time
 from distutils.util import strtobool
 
-import gym
+import gymnasium as gym
 import numpy as np
 import torch
 import torch.nn as nn
```
with stable-baselines3==2.0.0a5 and gymnasium==0.28.1

when i run 

`python3 cleanrl/cleanrl/dqn.py`

always after 'global_step=10009' execution stop with this error:

```global_step=10009, episodic_return=15.0
Traceback (most recent call last):
  File "/home/kris/dev/cleanRL-Gymnasium/cleanrl/cleanrl/dqn.py", line 195, in <module>
    data = rb.sample(args.batch_size)
  File "/home/kris/.local/lib/python3.9/site-packages/stable_baselines3/common/buffers.py", line 285, in sample
    return super().sample(batch_size=batch_size, env=env)
  File "/home/kris/.local/lib/python3.9/site-packages/stable_baselines3/common/buffers.py", line 110, in sample
    batch_inds = np.random.randint(0, upper_bound, size=batch_size)
  File "mtrand.pyx", line 765, in numpy.random.mtrand.RandomState.randint
  File "_bounded_integers.pyx", line 1247, in numpy.random._bounded_integers._rand_int64
ValueError: high <= 0
```

kvrban:
i think it was not intended to remove the following line:

`rb.add(obs, real_next_obs, actions, rewards, terminateds, infos)`

This could be the fix (fixed it for me):

```diff
diff --git a/cleanrl/dqn.py b/cleanrl/dqn.py
index 14864e7..4e73a6e 100644
--- a/cleanrl/dqn.py
+++ b/cleanrl/dqn.py
@@ -156,7 +156,7 @@ if __name__ == "__main__":
     start_time = time.time()
 
     # TRY NOT TO MODIFY: start the game
-    obs = envs.reset(seed=args.seed)
+    obs, _ = envs.reset(seed=args.seed)
     for global_step in range(args.total_timesteps):
         # ALGO LOGIC: put action logic here
         epsilon = linear_schedule(args.start_e, args.end_e, args.exploration_fraction * args.total_timesteps, global_step)
@@ -185,6 +185,7 @@ if __name__ == "__main__":
             for idx, d in enumerate(infos["_final_observation"]):
                 if d:
                     real_next_obs[idx] = infos["final_observation"][idx]
+        rb.add(obs, real_next_obs, actions, rewards, terminateds, infos)
 
         # TRY NOT TO MODIFY: CRUCIAL step easy to overlook
         obs = next_obs
```

pseudo-rnd-thoughts:
@kvrban Thanks for the comment but I think the plan is to complete this PR as several smaller PRs, see #370 and #371.

@arjun-kg or @vwxyzjn Should this PR be closed to avoid confusion?

vwxyzjn:
@pseudo-rnd-thoughts absolutely. Closing this PR now.

PR #288, opened 2022-10-05
Point offline RL users to tinkoff-ai/CORL

vwxyzjn:
## Description
https://github.com/tinkoff-ai/CORL seems to be a great offline RL library that shares a similar design philosophy with us. Since we are not really developing offline RL algorithms, we should give a friendly pointer to them, especially given that they have done the same for us:

image

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] New algorithm
- [x] Documentation

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ CleanRL is a Deep Reinforcement Learning library that provides high-quality sing
 
 You can read more about CleanRL in our [JMLR paper](https://www.jmlr.org/papers/volume23/21-1342/21-1342.pdf) and [documentation](https://docs.cleanrl.dev/).
 
-Good luck have fun :rocket:
+CleanRL only contains implementations of **online** deep reinforcement learning algorithms. If you are looking for **offline** algorithms, please check out [tinkoff-ai/CORL](https://github.com/tinkoff-ai/CORL), which shares a similar design philosophy as CleanRL.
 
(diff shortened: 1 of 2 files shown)


kinalmehta (approved):

dosssman (approved):

kinalmehta:
**Just a note:** I went through their reports and codes, and there is a significant performance gap between their scores and the paper's reported scores (CQL and DT are the ones I specifically checked). Though they are very transparent about that and have warnings in the readme and code files for which the performance does not match.

On a positive note, it is a good reference for anyone trying to understand paper through code.

vwxyzjn:
Merging now. Cc @Howuhh @Scitator @vkurenkov

Howuhh:
@kinalmehta Hi! Thank you for your time! 
Yeah, we know about underperformance for DT and CQL. 
I'm currently working on fixing the DT. It turned out to be a very small bug, because I concatenated the sequences incorrectly:
https://github.com/tinkoff-ai/CORL/blob/9069495537c9e22fb6a0c8abe7925f2bfa9b8a81/algorithms/dt.py#L312
After fixing, we are getting pretty close scores, so we just need to rerun all datasets, which take a bit of time.

CQL is a separate story, a magic algorithm that works differently for everyone (judging by the results of various publications). We also work on rewriting it from scratch again. However, we would be happy to see the contributors who knows CQL better than us.

kinalmehta:
Thank you for the update @Howuhh. It is indeed awesome work and the rigour with which you are evaluating the implementations is commendable.

It is good to know about EDAC's performance.

PR #293, opened 2022-10-14
Type hints

timoklein:
## Description
As discussed on Discord, I've done basic type hints for PPO and DQN. Everything checks out with mypy 0.982 (`mypy cleanrl/ppo.py --show-error-codes --ignore-missing-imports`). Of course we can have a discussion about whether that's the checker that will be used if we go further down the road of implementing this. I'll put comments in noteworthy places.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] New algorithm
- [ ] Documentation

## Checklist:

- [ ] Update poetry dependencies
- [ ] Type checker config.

### DIFF
--- a/cleanrl/dqn.py
+++ b/cleanrl/dqn.py
@@ -4,6 +4,7 @@ import os
 import random
 import time
 from distutils.util import strtobool
+from typing import Callable
 
 import gym
 import numpy as np
@@ -15,7 +16,7 @@ from stable_baselines3.common.buffers import ReplayBuffer
 from torch.utils.tensorboard import SummaryWriter
 
 
-def parse_args():
+def parse_args() -> argparse.Namespace:
     # fmt: off
     parser = argparse.ArgumentParser()
     parser.add_argument("--exp-name", type=str, default=os.path.basename(__file__).rstrip(".py"),
@@ -65,8 +66,8 @@ def parse_args():
     return args
 
 
-def make_env(env_id, seed, idx, capture_video, run_name):
-    def thunk():
+def make_env(env_id: str, seed: int, idx: int, capture_video: bool, run_name: str) -> Callable[[], gym.Env]:
+    def thunk() -> gym.Env:
         env = gym.make(env_id)
         env = gym.wrappers.RecordEpisodeStatistics(env)
         if capture_video:
@@ -82,7 +83,10 @@ def make_env(env_id, seed, idx, capture_video, run_name):
 
 # ALGO LOGIC: initialize agent here:
 class QNetwork(nn.Module):
-    def __init__(self, env):
+
+    network: nn.Sequential
+
+    def __init__(self, env: gym.vector.SyncVectorEnv):
         super().__init__()
         self.network = nn.Sequential(
             nn.Linear(np.array(env.single_observation_space.shape).prod(), 120),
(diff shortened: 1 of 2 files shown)


cleanrl/ppo.py
```diff
@@ -159,15 +166,24 @@ def get_action_and_value(self, x, action=None):
     # env setup
     envs = gym.vector.SyncVectorEnv(
         [make_env(args.env_id, args.seed + i, i, args.capture_video, run_name) for i in range(args.num_envs)]
-    )
+    )  # type:ignore[abstract]
```
timoklein, line 169: SyncVectorEnv inherits from VectorEnv which inherits from Env. For older gym version (I'm currently on 0.23.1), Env is an ABC with abstract method `render` that is not overriden by any of the vector envs. Since it's fixed in the newest gym release and not our issue, I'm ignoring here.

cleanrl/ppo.py
```diff
@@ -159,15 +166,24 @@ def get_action_and_value(self, x, action=None):
...
     assert isinstance(envs.single_action_space, gym.spaces.Discrete), "only discrete action space is supported"
+    # Handling gym shapes being Optionals (variant 1)
+    # Personally i'd prefer the asserts
+    assert isinstance(envs.single_observation_space.shape, tuple), "shape of observation space must be defined"
+    assert isinstance(envs.single_action_space.shape, tuple), "shape of action space must be defined"
+
+    # Handling gym shapes being Optionals (variant 2)
+    # Once could also cast inside each call but in my eyes that's not conducive to readability
+    obs_space_shape = cast(tuple[int, ...], envs.single_observation_space.shape)
+    action_space_shape = cast(tuple[int, ...], envs.single_action_space.shape)
```
timoklein, line 179: Gym spaces can in theory return `None` shapes. Mypy will complain about this when concatenating the shape tuples later.
- Option 1 is to use a cast either once here or every time the spaces are accessed. I don't think that's very readable.
- Option 2 is to assert that the space shapes are tuples. Doing it once here fixes all errors for the rest of the code. Since there's an assert in this place already anyway I think this is the better option.

cleanrl/dqn.py
```diff
@@ -92,11 +96,11 @@ def __init__(self, env):
             nn.Linear(84, env.single_action_space.n),
         )
 
-    def forward(self, x):
+    def forward(self, x: torch.FloatTensor) -> torch.FloatTensor:
```
timoklein, line 99: I've used FloatTensor here but we should just use `torch.Tensor` if type hints are pursued further.

vwxyzjn (commented):
Thank you! The PR looks good. I have left some comments

cleanrl/ppo.py
Daraan: Adding `from __future__ import annotation` at the top would guarantee the 3.7+ compatibility, and make dropping support later on quite easy.

cleanrl/ppo.py
```diff
@@ -116,10 +121,12 @@ def __init__(self, envs):
...
         )
 
-    def get_value(self, x):
+    def get_value(self, x: torch.Tensor) -> torch.Tensor:
         return self.critic(x)
 
-    def get_action_and_value(self, x, action=None):
+    def get_action_and_value(
+        self, x: torch.Tensor, action: Optional[torch.Tensor] = None
+    ) -> tuple[torch.Tensor, torch.Tensor, torch.Tensor, torch.Tensor]:
```
Daraan, line 129: Adding `from __future__ import annotations` at the top would guarantee the 3.7+ compatibility, and make dropping support later on quite easy.

PR #323, opened 2022-11-22
Fix DQN target update frequency

### LINKED ISSUE #322: Target network isn't updated to the correct frequency when `target_network_frequency % train_frequency != 0`
Because this `if` statement

https://github.com/vwxyzjn/cleanrl/blob/c37a3ec4ef8d33ab7c8a69d4d2714e3817739365/cleanrl/dqn.py#L205

is inside this one

https://github.com/vwxyzjn/cleanrl/blob/c37a3ec4ef8d33ab7c8a69d4d2714e3817739365/cleanrl/dqn.py#L185

Consequently, the target network is updated when `global_step % train_frequency == 0 and global_step % target_network_frequency == 0`.

For example, when you run

```shell
python cleanrl/dqn.py --target-network-frequency 501
```

The target network is updated every 5010 timesteps, not every 501 timesteps.

qgallouedec:
## Description

Closes #322 

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] New algorithm
- [ ] Documentation

## Checklist:

### DIFF
--- a/cleanrl/dqn.py
+++ b/cleanrl/dqn.py
@@ -182,28 +182,29 @@ if __name__ == "__main__":
         obs = next_obs
 
         # ALGO LOGIC: training.
-        if global_step > args.learning_starts and global_step % args.train_frequency == 0:
-            data = rb.sample(args.batch_size)
-            with torch.no_grad():
-                target_max, _ = target_network(data.next_observations).max(dim=1)
-                td_target = data.rewards.flatten() + args.gamma * target_max * (1 - data.dones.flatten())
-            old_val = q_network(data.observations).gather(1, data.actions).squeeze()
-            loss = F.mse_loss(td_target, old_val)
+        if global_step > args.learning_starts:
+            if global_step % args.train_frequency == 0:
+                data = rb.sample(args.batch_size)
+                with torch.no_grad():
+                    target_max, _ = target_network(data.next_observations).max(dim=1)
+                    td_target = data.rewards.flatten() + args.gamma * target_max * (1 - data.dones.flatten())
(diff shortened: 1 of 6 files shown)


cleanrl/dqn.py
```diff
@@ -201,9 +201,9 @@ def linear_schedule(start_e: float, end_e: float, duration: int, t: int):
             loss.backward()
             optimizer.step()
 
-            # update the target network
-            if global_step % args.target_network_frequency == 0:
-                target_network.load_state_dict(q_network.state_dict())
+        # update the target network
+        if global_step % args.target_network_frequency == 0:
+            target_network.load_state_dict(q_network.state_dict())
```
vwxyzjn, line 206: Maybe change it to 

```
        # update the target network
        if global_step > args.learning_starts and global_step % args.target_network_frequency == 0:
            target_network.load_state_dict(q_network.state_dict())
```

vwxyzjn (changes requested):
@qgallouedec thanks so much for the fix!

cleanrl/dqn.py
qgallouedec: What about something like this instead:

```python
if global_step > learning_starts :
    if global_step % train_frequency == 0:
        ...

    if global_step % target_network_frequency == 0:
        ...
```

I find it more readable.

vwxyzjn: That would work great.

vwxyzjn (approved):
LGTM. Given that this change does not impact the performance of our algorithm variants using their default parameters, we do not need to re-run the benchmark for these variants.

PR #334, opened 2022-12-28
chore: simplify syntax

vwxyzjn:
## Description
This PR does a simple refactoring on the dataclass syntax in `ppo_atari_envpool_xla_jax_scan.py`.

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] New algorithm
- [ ] Documentation
- [x] minor refactor

PR #337, opened 2022-12-31
bug: incorrect logic in GAE calculation

vwxyzjn:
## Description
#334 actually introduced a major bug... It's totally my bad 🙈

```
        storage = storage.replace(
            advantages=advantages,
            returns=storage.advantages + storage.values,
        )
```
should have been the following because `storage.advantages` has zero values.

```
        storage = storage.replace(
            advantages=advantages,
            returns=advantages + storage.values,
        )
```

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] New algorithm
- [ ] Documentation


51616 (commented):
This looks correct to me. Btw, did the gae test in #334 run correctly? Seems like the bug is in the test file as well.

PR #385, opened 2023-05-09
Update wandb to 0.13.11

pseudo-rnd-thoughts:
## Description

The indention of following code is inconsistent between files, I am assuming that `ddpg_contiuous_action` is correct and a fixing these but it might be the other way around
```python
    if args.track and args.capture_video:
        for filename in os.listdir(f"videos/{run_name}"):
            if filename not in video_filenames and filename.endswith(".mp4"):
                wandb.log({f"videos": wandb.Video(f"videos/{run_name}/{filename}")})
                video_filenames.add(filename)
```

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] New algorithm
- [ ] Documentation

## Checklist:

    - For new feature or bug fix:
    - For new algorithm:


cleanrl/ppo_continuous_action.py
```diff
@@ -322,11 +321,12 @@ def get_action_and_value(self, x, action=None):
...
 
-        if args.track and args.capture_video:
-            for filename in os.listdir(f"videos/{run_name}"):
-                if filename not in video_filenames and filename.endswith(".mp4"):
-                    wandb.log({f"videos": wandb.Video(f"videos/{run_name}/{filename}")})
-                    video_filenames.add(filename)
-
     envs.close()
+
+    video_filenames = set()
```
vwxyzjn, line 326: This should be at the beginning of the training loop

pseudo-rnd-thoughts:
@vwxyzjn I realised that the easier solution is to update wandb to 0.13.11 with https://github.com/wandb/wandb/pull/5008 to support gymnasium rather than adding the code to each of the implementations

vwxyzjn:
> @vwxyzjn I realised that the easier solution is to update wandb to 0.13.11 with [wandb/wandb#5008](https://github.com/wandb/wandb/pull/5008) to support gymnasium rather than adding the code to each of the implementations

It probably won’t work because we have both gymnasium and gym installed.

sdpkjc:
https://github.com/vwxyzjn/cleanrl/blob/e19f85819253f2952f8af6b09ea393e5e16cf891/requirements/requirements-atari.txt#L73

In fact, our code is already using wandb==0.13.11, we need to update it officially in the `pyproject.toml` file and run `poetry lock`.

https://github.com/vwxyzjn/cleanrl/blob/e19f85819253f2952f8af6b09ea393e5e16cf891/cleanrl/dqn.py#L137

Current `dqn.py` has updated the way it records videos, and it still works. (I've tested it.) Because wandb preferentially uses the gymnasium library.

However, the current part of the code that uses gym may have bugs due to wandb updates, which should be fixed by setting `_gym_version_lt_0_26=True`.

https://github.com/wandb/wandb/blob/134cf1c69200b7211167a50455e32642e1089fcf/wandb/integration/gym/__init__.py#L43-L86

vwxyzjn:
@sdpkjc thanks for the comment. @pseudo-rnd-thoughts I see what’s the issue. Currently, because wandb version in the lock file is already 0.13.11 and its `monitor_gym` will work for gymnasium scripts. I also confirmed that `dqn.py` works https://wandb.ai/costa-huang/cleanRL/runs/mco6owqv?workspace=user-costa-huang.

However, `ppo.py` with `--capture-video` no longer works, because it still uses `gym` and now `monitor_gym` only works with gymnasium. 

Let's just ignore the capture video problem in `ppo.py` and continue our gymnasium migration. @pseudo-rnd-thoughts can you update the lock file in this PR with `poetry lock --no-update`?  We can merge after that.

vwxyzjn (approved):
This also LGTM. Feel free to merge after #386. Thanks so much @pseudo-rnd-thoughts

PR #392, opened 2023-05-23
Adding MPO and DMPO

Jogima-cyber:
## Description

I've started investigating the MPO algorithm family and I wish to do it in a clean RL fashion (benchmarking everything and having one file implementation for each algorithm) as it seems to me to be a good way to make RL algorithms accessible to everyone.

Apart from deepmind's official implementation, which is hard to use or just analyze because they built their library in a modular fashion and rely heavily on calls to other libraries that they've developed and that are used only by them, there are very few resources on these algorithms, and none when it comes to trustworthy resources.

Moreover this family lacks a real benchmark, since in the related papers they only do benchmarking for a segment of algorithms of this family (TD(5) and Retrace, whereas we'd like to see benchmarks for the use of a distributional critic) and this benchmarking is sparse (for the gym mujoco envs it doesn't include all envs, was done on v1, and doesn't compare results with other algorithms than a version of SAC)

Nonetheless I think this family should be thoughtfully investigated because of the following claims in the robotics robotics continuous control domain:
Has a way better sample efficiency than PPO and is as much insensible to hyperparameters tuning as PPO is (the later claim is very important as practitioners in the robotic field usually cannot make DDPG/TD3/SAC work on real robots because of the need for these algorithms to tune a lot the hyperparameters)
Has same sample efficiency as SAC but better asymptotic results
Furthermore the repeated use by deepmind of this family of algorithms in robotics for 5 years, and recently in the quite impressive https://arxiv.org/pdf/2304.13653.pdf paper, may be a signal that this family is actually a very good family of algorithms for robotic continuous control.

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [x] New algorithm
- [ ] Documentation

## Checklist:

    - For new feature or bug fix:
    - For new algorithm:

### DIFF
--- a/cleanrl/dmpo_continuous_action.py
+++ b/cleanrl/dmpo_continuous_action.py
@@ -1,0 +1,652 @@
+import argparse
+import os
+import random
+import time
+from distutils.util import strtobool
+
+import gym
+import numpy as np
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+import torch.optim as optim
+from stable_baselines3.common.buffers import ReplayBuffer
+from torch.utils.tensorboard import SummaryWriter
+from typing import NamedTuple, Any, Dict, Generator, List, Optional, Union
+from gym import spaces
+from stable_baselines3.common.vec_env import VecNormalize
+
+try:
+    # Check memory used by replay buffer when possible
+    import psutil
+except ImportError:
+    psutil = None
+
+# MPO with C51 style critic loss (critic outputs categorical distribution of the q function,
+# instead of just mean of this distribution, this is called "distributional RL")
+
+def parse_args():
+    # fmt: off
+    parser = argparse.ArgumentParser()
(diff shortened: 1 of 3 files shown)


Jogima-cyber:
Who should run the benchmarks? And how? Because we'd like to check if we get the same as the acme benchmarks:
deepminds_official_benchmarks

but the evaluation is done differently than cleanrl and training is mostly for 1e7 steps (in clean RL benchmarks this is usually 1e6 if I'm not mistaken).

vwxyzjn:
If it’s possible, you should run the benchmark. Regarding experiment settings, perfect replication is difficult (e.g., do we know their PPO settings?). It’s up to you if you want to use 1e6 or 1e7.

Jogima-cyber:
Okay, I'm gonna run the benchmark!

Jogima-cyber:
Results obtained with the proposed DMPO.
compare

PR #370, opened 2023-04-01
Add gymnasium support for DQN

valentin-cnt:
## Description
This PR updates the DQN files to the lastest version of gymnasium, replacing gym.

- [X] `dqn.py`
- [X] `dqn_jax.py`
- [X] `dqn_atari.py`
- [X] `dqn_atari_jax.py`

## Types of changes

- [ ] Bug fix
- [X] New feature
- [ ] New algorithm
- [ ] Documentation

## Checklist:

- [X] I've read the [CONTRIBUTION](https://docs.cleanrl.dev/contribution/) guide (**required**).
- [X] I have ensured `pre-commit run --all-files` passes (**required**).
- [x] I have updated the tests accordingly (if applicable).
- [x] I have updated the documentation and previewed the changes via `mkdocs serve`.
    - [x] I have explained note-worthy implementation details.
    - [x] I have explained the logged metrics.
    - [x] I have added links to the original paper and related papers.

If you need to run benchmark experiments for a performance-impacting changes:

- [X] I have contacted @vwxyzjn to obtain access to the [openrlbenchmark W&B team](https://wandb.ai/openrlbenchmark).
- [X] I have used the [benchmark utility](/get-started/benchmark-utility/) to submit the tracked experiments to the [openrlbenchmark/cleanrl](https://wandb.ai/openrlbenchmark/cleanrl) W&B project, optionally with `--capture-video`.
- [x] I have performed RLops with `python -m openrlbenchmark.rlops`.
    - For new feature or bug fix:
        - [x] I have used the RLops utility to understand the performance impact of the changes and confirmed there is no regression.
    - For new algorithm:
        - [x] I have created a table comparing my results against those from reputable sources (i.e., the original paper or other reference implementation).
    - [x] I have added the learning curves generated by the `python -m openrlbenchmark.rlops` utility to the documentation.
    - [x] I have added links to the tracked experiments in W&B, generated by `python -m openrlbenchmark.rlops ....your_args... --report`,  to the documentation.

## Regression report

```
python -m openrlbenchmark.rlops \
    --filters '?we=openrlbenchmark&wpn=cleanrl&ceik=env_id&cen=exp_name&metric=charts/episodic_return' \
        'dqn_atari_jax?tag=rlops-pilot' \
        'dqn_atari_jax?tag=pr-370-atari-jax' \
    --env-ids Breakout-v5 BeamRider-v5 Pong-v5 \
    --check-empty-runs False \
    --ncols 5 \
    --ncols-legend 2 \
[...]

PR #395, opened 2023-06-06
handle num_envs > 1 in DQN

ronuchit:
## Description
We remove the requirement that `num_envs=1` in the DQN implementation. See discussion in:
https://github.com/vwxyzjn/cleanrl/pull/370#issuecomment-1577698962

## Types of changes
- [ x ] Bug fix
- [ ] New feature
- [ ] New algorithm
- [ ] Documentation

## Checklist:
- [ x ] I've read the [CONTRIBUTION](https://docs.cleanrl.dev/contribution/) guide (**required**).

### DIFF
--- a/cleanrl/dqn.py
+++ b/cleanrl/dqn.py
@@ -72,7 +72,6 @@ def parse_args():
         help="the frequency of training")
     args = parser.parse_args()
     # fmt: on
-    assert args.num_envs == 1, "vectorized envs are not supported at the moment"
 
     return args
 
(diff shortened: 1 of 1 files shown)


sdpkjc:
Here we may need to pay attention to the problem of the increment of `global_step` (`+1`?  or `+args.num_envs`?)

- If `+args.num_envs`, then we need to consider the divisibility relationship between `global_step` and log recording points (for example: `if global_step % 100 == 0:`), and the divisibility relationship between `global_step` and `args.train_frequency` (`if global_step % args.train_frequency == 0:`)
  - We should give some notes or warnings when we cannot be divided.
- If `+1`, then we need to consider whether our `total_timesteps` is still correct.  We may need to write the log like this: `writer.add_scalar("charts/episodic_return", info["episode"]["r"], global_step*args.num_envs)`.

PR #408, opened 2023-07-12
Update ppo_pettingzoo_ma_atari.py

elliottower:
This PR updates the pettingzoo multiagent atari example to use gymnasium rather than gym and to use the current pettingzoo API (with termination and truncation, following gymnasium/gym v26). I've had some people ask about more in-depth CleanRL resources for PettingZoo, so I figure updating this would be a good start. 

Unfortunately it seems like the record episode statistics won't work because it expects a dict for info, and the supersuit's `concat_vec_env` makes the info into a list of dicts. Could write a custom wrapper to do that but it's not entirely clear to me that's the best way to do things. We have been looking into mirroring the gymnasium vector API into PettingZoo which would allow the recorder class and most other gymnasium functionality to work, as far as I can tell, but that will take some time. 

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] New algorithm
- [ ] Documentation

## Checklist:

    - For new feature or bug fix:
    - For new algorithm:

### DIFF
--- a/cleanrl/ppo_pettingzoo_ma_atari.py
+++ b/cleanrl/ppo_pettingzoo_ma_atari.py
@@ -6,7 +6,7 @@ import random
 import time
 from distutils.util import strtobool
 
-import gym
+import gymnasium as gym
 import numpy as np
 import supersuit as ss
 import torch
@@ -156,11 +156,10 @@ if __name__ == "__main__":
     env = ss.frame_stack_v1(env, 4)
     env = ss.agent_indicator_v0(env, type_only=False)
     env = ss.pettingzoo_env_to_vec_env_v1(env)
-    envs = ss.concat_vec_envs_v1(env, args.num_envs // 2, num_cpus=0, base_class="gym")
+    envs = ss.concat_vec_envs_v1(env, args.num_envs // 2, num_cpus=0, base_class="gymnasium")
     envs.single_observation_space = envs.observation_space
     envs.single_action_space = envs.action_space
     envs.is_vector_env = True
-    envs = gym.wrappers.RecordEpisodeStatistics(envs)
     if args.capture_video:
         envs = gym.wrappers.RecordVideo(envs, f"videos/{run_name}")
     assert isinstance(envs.single_action_space, gym.spaces.Discrete), "only discrete action space is supported"
@@ -173,14 +172,17 @@ if __name__ == "__main__":
     actions = torch.zeros((args.num_steps, args.num_envs) + envs.single_action_space.shape).to(device)
     logprobs = torch.zeros((args.num_steps, args.num_envs)).to(device)
     rewards = torch.zeros((args.num_steps, args.num_envs)).to(device)
-    dones = torch.zeros((args.num_steps, args.num_envs)).to(device)
+    terminations = torch.zeros((args.num_steps, args.num_envs)).to(device)
+    truncations = torch.zeros((args.num_steps, args.num_envs)).to(device)
     values = torch.zeros((args.num_steps, args.num_envs)).to(device)
 
     # TRY NOT TO MODIFY: start the game
     global_step = 0
     start_time = time.time()
-    next_obs = torch.Tensor(envs.reset()).to(device)
-    next_done = torch.zeros(args.num_envs).to(device)
+    next_obs, info = envs.reset(seed=args.seed)
+    next_obs = torch.Tensor(next_obs).to(device)
+    next_termination = torch.zeros(args.num_envs).to(device)
+    next_truncation = torch.zeros(args.num_envs).to(device)
     num_updates = args.total_timesteps // args.batch_size
 
     for update in range(1, num_updates + 1):
@@ -193,7 +195,8 @@ if __name__ == "__main__":
         for step in range(0, args.num_steps):
             global_step += 1 * args.num_envs
             obs[step] = next_obs
-            dones[step] = next_done
+            terminations[step] = next_termination
+            truncations[step] = next_truncation
 
             # ALGO LOGIC: action logic
             with torch.no_grad():
@@ -203,10 +206,11 @@ if __name__ == "__main__":
             logprobs[step] = logprob
 
             # TRY NOT TO MODIFY: execute the game and log data.
-            next_obs, reward, done, info = envs.step(action.cpu().numpy())
+            next_obs, reward, termination, truncation, info = envs.step(action.cpu().numpy())
             rewards[step] = torch.tensor(reward).to(device).view(-1)
-            next_obs, next_done = torch.Tensor(next_obs).to(device), torch.Tensor(done).to(device)
+            next_obs, next_termination, next_truncation = torch.Tensor(next_obs).to(device), torch.Tensor(termination).to(device), torch.Tensor(truncation).to(device)
 
+            # TODO: fix this
             for idx, item in enumerate(info):
                 player_idx = idx % 2
                 if "episode" in item.keys():
@@ -219,6 +223,8 @@ if __name__ == "__main__":
             next_value = agent.get_value(next_obs).reshape(1, -1)
             advantages = torch.zeros_like(rewards).to(device)
             lastgaelam = 0
+            next_done = torch.maximum(next_termination, next_truncation)
+            dones = torch.maximum(terminations, truncations)
             for t in reversed(range(args.num_steps)):
                 if t == args.num_steps - 1:
                     nextnonterminal = 1.0 - next_done


elliottower:
Not sure how big of a problem it is not being able to record episode statistics, it may be better to keep things as they are currently so it doesn't lose functionality, and I can put this updated version onto the pettingzoo docs. 

There may be a way to get the episode statistics to work but it seems difficult as I explained above

elliottower:
@vwxyzjn I'm imagining this would also require running the benchmarks? Let me know if you have any thoughts on this, I've also been considering doing an action masking example for other PZ envs, would that be something you're interested in having here?

cleanrl/ppo_pettingzoo_ma_atari.py
```diff
@@ -219,6 +227,8 @@ def get_action_and_value(self, x, action=None):
             next_value = agent.get_value(next_obs).reshape(1, -1)
             advantages = torch.zeros_like(rewards).to(device)
             lastgaelam = 0
+            next_done = torch.maximum(next_termination, next_truncation)
```
vwxyzjn, line 230: Could you do https://github.com/vwxyzjn/cleanrl/blob/f62ad1f942890db7b752f4ad79b65be53917d3c2/cleanrl/ppo_continuous_action.py#L213 to be consistent?

vwxyzjn (commented):
The changes looks good to me. I think we should do a benchmark run though because pettingzoo’s version changed a lot. Having an action mask example would be good too, but let’s do that in a separate PR

cleanrl/ppo_pettingzoo_ma_atari.py
elliottower: Do you want it specifically to be numpy rather than torch as well? I'm assuming for the sake of speed it's simpler to keep things on device so I went with `torch.logical_or`, but I can change it to np if that's better for some reason

vwxyzjn: The motivation is to *minimize the lines of code difference* between `ppo_pettingzoo_ma_atari.py` and `ppo_atari.py`. if `torch.logical_or` does not give significant end-to-end speed up, it's more important to be consistent.

elliottower: Makes sense, but there seem to be issues with most of the other options (writing it out this way so it's a bit less confusing)

1. `np.logical_or(next_termination, next truncation)`,  CI but testing on my desktop with CUDA it has an error due to different devices: `TypeError: can't convert cuda:0 device type tensor to numpy. Use Tensor.cpu() to copy the tensor to host memory first.`. 

2. np.logical_or(next_termination.cpu(), next_truncation.cpu() works for that line but fails later on:
`Traceback (most recent call last):
  File "/mnt/d/GitHub/PettingZoo/tutorials/CleanRL/cleanrl_advanced.py", line 267, in <module>
    rewards[t] + args.gamma * nextvalues * nextnonterminal - values[t]
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
`, where `nextnonterminal` is on CPU. 
3.  `np.logical_or()` requires the two args to be on CPU, I could convert them and then back to device, but that seems messy and inefficient: `np.logical_or(next_termination.cpu(), next_truncation.cpu()).to(device)`. 
4. `torch.logical_or(next_termination, next_truncation)` works for that line but fails when trying to do subtraction on the line below (`1.0 - next_done`), as bool tensors don't support it: ```RuntimeError: Subtraction, the `-` operator, with a bool tensor is not supported. If you are trying to invert a mask, use the `~` or `logical_not()` operator instead.```
   * Another option is to then cast it to an int array, but that doesn't seem the most concise either, especially when it's the exact same behavior as maximum.

Given this. l think my original choice `torch.maximum(next_termination, next_truncation)` makes the most sense. I also looked through the other code to find uses of `np.logical_or()` and couldn't find any. Seems bizarre to suddenly take the arrays off the CUDA device midway through the training loop anyways, but ultimately it's obviously your call.

cleanrl/ppo_pettingzoo_ma_atari.py
```diff
@@ -219,6 +227,8 @@ def get_action_and_value(self, x, action=None):
             next_value = agent.get_value(next_obs).reshape(1, -1)
             advantages = torch.zeros_like(rewards).to(device)
             lastgaelam = 0
+            next_done = np.logical_or(next_termination, next_truncation)
```
KaleabTessera, line 230: I think there is bug here. We should still bootstrap if `next_truncation=True` :

> "you should bootstrap if infos[env_idx]["TimeLimit.truncated"] is True (episode over due to a timeout/truncation) or dones[env_idx] is False (episode not finished)." - [stable baselines](https://stable-baselines3.readthedocs.io/en/master/guide/vec_envs.html)

So `next_done=next_termination` and  `dones=terminations` (probs just use next_terminations and terminations directly). 

To implement this correctly we also need access to `terminal_observation` from `pettingzoo_env_to_vec_env_v1`. I have a [PR](https://github.com/Farama-Foundation/SuperSuit/pull/233) out for this. Then we can implement something like [this](https://github.com/DLR-RM/stable-baselines3/blob/69afefc91d408d352b4224ae5244ad2c32bb7634/stable_baselines3/common/on_policy_algorithm.py#L212) to do correct bootstrapping for truncating/timeout.

KaleabTessera (changes requested):
I think there is possibly a bug here. Truncation should be treated differently to termination when bootstrapping - as done in stable baselines. This made a pretty big diff in my own experiments.

cleanrl/ppo_pettingzoo_ma_atari.py
elliottower: Good catch @KaleabTessera would you be willing to update this branch with the changes? I can give you edit access, I currently have a lot of other obligations from work so don’t have much time for this

elliottower: Oh shoot it’s a patch-1 so I don’t know if you can be given access. But if you clone the repo you can make a new branch from this branch and make a new PR if it’s not possible to edit this branch? Or maybe make a PR to update this branch itself. Sorry I can’t help more

vwxyzjn: FYI I am doing a refactor at #424 . Gonna try run a whole suite of benchmark soon.

ezhang7423:
Are there any updates on this?

elliottower:
Looks like https://github.com/vwxyzjn/cleanrl/pull/424 was merged, 

> Are there any updates on this?

Looks like https://github.com/vwxyzjn/cleanrl/pull/424 was merged, but it didn't update the PettingZoo example besides a minor CLI arguments change. Re-reading Costa's messages I think I misinterpreted him that he was intending to update this himself (or maybe he didn't have time). 

Anyways, I have some time today and will try to resolve these conflicts and integrate @KaleabTessera's suggestions, so we can at least have an updated version of this script. Won't have time for benchmarking in the near future but could eventually get to it.

vwxyzjn:
Yeah sorry @elliottower things have gotten busy. Feel free to submit a PR. As long as you can reproduce the existing benchmark experiments we can merge :)

elliottower:
No worries, sounds good. I'll just use this same PR for simplicity's sake.

elliottower:
Btw just FYI there's a bunch of already merged branches in this repo which could probably deleted (am pulling the most recent master branch and see a huge list)

cleanrl/ppo_pettingzoo_ma_atari.py
elliottower: > I think there is bug here. We should still bootstrap if `next_truncation=True` :
> 
> > "you should bootstrap if infos[env_idx]["TimeLimit.truncated"] is True (episode over due to a timeout/truncation) or dones[env_idx] is False (episode not finished)." - [stable baselines](https://stable-baselines3.readthedocs.io/en/master/guide/vec_envs.html)
> 
> So `next_done=next_termination` and `dones=terminations` (probs just use next_terminations and terminations directly e.g. `nextnonterminal = 1.0 - next_termination` ).
> 
> To implement this correctly we also need access to `terminal_observation` from `pettingzoo_env_to_vec_env_v1` since we need access to the true terminal obs and not the obs returned by the next restart (the case currently -- so we need infos to provide access to the terminal obs). I have a [PR](https://github.com/Farama-Foundation/SuperSuit/pull/233) out for this . Then we can implement something like [this](https://github.com/DLR-RM/stable-baselines3/blob/69afefc91d408d352b4224ae5244ad2c32bb7634/stable_baselines3/common/on_policy_algorithm.py#L212) to do correct bootstrapping for truncating/timeout.

Btw, just as an update, the SuperSuit PR linked has been merged. My only concern with this is that whatever bootstrapping behavior is done here should mirror what is done with the single agent PPO implementations, so this is a question for @vwxyzjn. 

My inclination is to keep the logic as it currently is in this PR and address that bootstrapping issue in another PR (maybe @KaleabTessera is interested in doing that? I don't have a whole lot of time to look into it nor am I the best person to do it as I'm not an expert)

Issue #198, opened 2022-06-10
PPO timeout proper handling 


Howuhh:
Hi! I'm a bit puzzled as to how a timeout could be handled correctly in your implementation of PPO (well, this is relevant for all variants). I am especially surprised by envpool, because seems like they do not return the last real state at all (like gym vec envs do in `info`). 

Were there any ideas on how to do it right? I'd like to make a PR with that fix, but I haven't figured out how yet. Buffer in PPO is a bit more confusing than in off-policy algos where there is a clear (s, s', d) thing..

vwxyzjn:
Proper timeout handling is definitely an interesting issue I want to look into further. I think the best resource is https://github.com/DLR-RM/stable-baselines3/pull/658. Happy to take a look into this together :) 

If you are submitting a fix PR, I highly suggest doing it just on `ppo.py` to show a proof-of-concept before applying the changes to other PPO variants.

Howuhh:
Thanks for the link! I'll take a look and see what I can come up with.

Howuhh:
After some thinking and sketching on a piece of paper, it seems to me that it could be solved this way (just proposal for now):
```python
# define buffer's here: states, actions, rewards, dones, values, logprobs
# Initial obs and info
state, prev_info = torch.Tensor(envs.reset()).to(device), {}

for step in range(0, args.num_steps):
    with torch.no_grad():
        action, logprob, _, value = agent.get_action_and_value(state)
    next_state, reward, done, info = envs.step(action.cpu().numpy())

    # full transition on step T
    states[step] = state
    actions[step] = action
    rewards[step] = torch.tensor(reward).to(device).view(-1)
    dones[step] = done
    # also PPO stuff for step T
    logprobs[step] = logprob
    values[step] = value.flatten()

    # Here we should check for timeout on previous step
    if step > 1 and torch.any(dones[step - 1]):
        # if on prev step was timeout, then we should
        # 1. set dones[step - 1][env_id] to False, as it was not real done
        # 2. set values[step][env_id] to V(prev_info['terminal_observation']) (as it is real next_state for previous step)
        for env_id in range(args.num_envs):
            timeout = "TimeLimit.truncated" in prev_info[env_id] and prev_info[env_id]["TimeLimit.truncated"]
            if timeout:
                terminal_state = torch.tensor(prev_info[env_id]["terminal_observation"], device=device)
                # Set done to false as it was timeout
                dones[step - 1][env_id] = 0 
[...]

PR #448, opened 2024-02-29
Enhancing Termination and Truncation Handling in CleanRL's PPO Algorithm

Josh00-Lu:
## Description
I have noticed that the issues regarding termination and truncation in the PPO (Proximal Policy Optimization) algorithm have not been addressed and resolved. Therefore, I have incorporated relevant handling methods. Specifically, for a vector environment (vec env), when termination or truncation is received at the current time step, the actual next state will be provided in the `"final_observation"`, and we need to record this information. Also, here, `done = termination or truncation`.

On the other hand, when calculating delta and advantage using GAE (Generalized Advantage Estimation), delta should be computed using `termination`, while advantages should be computed with `done`. For more details, please refer to the PR (Pull Request) codes and comments I have provided.

Notes: 
- [x] Since we only next `dones[t+1]` and `terminations[t+1]` when computing advantages using GAE, I have make slight code adjustments to simplify the code logics.
- [x] Now we need to store `next_terminations` and `next_obs`.

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] New algorithm
- [ ] Documentation

## Checklist:

    - For new feature or bug fix:
    - For new algorithm:

### DIFF
--- a/cleanrl/ppo_continuous_action.py
+++ b/cleanrl/ppo_continuous_action.py
@@ -184,18 +184,21 @@ if __name__ == "__main__":
 
     # ALGO Logic: Storage setup
     obs = torch.zeros((args.num_steps, args.num_envs) + envs.single_observation_space.shape).to(device)
+    next_obs = torch.zeros((args.num_steps, args.num_envs) + envs.single_observation_space.shape).to(device)
     actions = torch.zeros((args.num_steps, args.num_envs) + envs.single_action_space.shape).to(device)
     logprobs = torch.zeros((args.num_steps, args.num_envs)).to(device)
     rewards = torch.zeros((args.num_steps, args.num_envs)).to(device)
-    dones = torch.zeros((args.num_steps, args.num_envs)).to(device)
+    next_dones = torch.zeros((args.num_steps, args.num_envs)).to(device)
+    next_terminations = torch.zeros((args.num_steps, args.num_envs)).to(device)
     values = torch.zeros((args.num_steps, args.num_envs)).to(device)
 
     # TRY NOT TO MODIFY: start the game
     global_step = 0
     start_time = time.time()
-    next_obs, _ = envs.reset(seed=args.seed)
-    next_obs = torch.Tensor(next_obs).to(device)
+    next_ob, _ = envs.reset(seed=args.seed)
+    next_ob = torch.Tensor(next_ob).to(device)
     next_done = torch.zeros(args.num_envs).to(device)
+    next_termination = torch.zeros(args.num_envs).to(device)
 
     for iteration in range(1, args.num_iterations + 1):
         # Annealing the rate if instructed to do so.
@@ -206,24 +209,34 @@ if __name__ == "__main__":
 
         for step in range(0, args.num_steps):
             global_step += args.num_envs
-            obs[step] = next_obs
-            dones[step] = next_done
-
+            
+            ob = next_ob
             # ALGO LOGIC: action logic
             with torch.no_grad():
-                action, logprob, _, value = agent.get_action_and_value(next_obs)
-                values[step] = value.flatten()
-            actions[step] = action
-            logprobs[step] = logprob
+                action, logprob, _, value = agent.get_action_and_value(ob)
 
             # TRY NOT TO MODIFY: execute the game and log data.
-            next_obs, reward, terminations, truncations, infos = envs.step(action.cpu().numpy())
-            next_done = np.logical_or(terminations, truncations)
+            next_ob, reward, next_termination, next_truncation, info = envs.step(action.cpu().numpy())
+            
+            # Correct next obervation (for vec gym)
+            real_next_ob = next_ob.copy()
+            for idx, trunc in enumerate(next_truncation):
+                if trunc:
+                    real_next_ob[idx] = info["final_observation"][idx]
+            next_ob = torch.Tensor(next_ob).to(device)
+            
+            # Collect trajectory
+            obs[step] = torch.Tensor(ob).to(device)
+            next_obs[step] = torch.Tensor(real_next_ob).to(device)
+            actions[step] = torch.Tensor(action).to(device)
+            logprobs[step] = torch.Tensor(logprob).to(device)
+            values[step] = torch.Tensor(value.flatten()).to(device)
+            next_terminations[step] = torch.Tensor(next_termination).to(device)
+            next_dones[step] = torch.Tensor(np.logical_or(next_termination, next_truncation)).to(device)
             rewards[step] = torch.tensor(reward).to(device).view(-1)
-            next_obs, next_done = torch.Tensor(next_obs).to(device), torch.Tensor(next_done).to(device)
-
-            if "final_info" in infos:
-                for info in infos["final_info"]:
+            
+            if "final_info" in info:
+                for info in info["final_info"]:
                     if info and "episode" in info:
                         print(f"global_step={global_step}, episodic_return={info['episode']['r']}")
                         writer.add_scalar("charts/episodic_return", info["episode"]["r"], global_step)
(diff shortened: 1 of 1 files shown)


vwxyzjn:
Hi @Josh00-Lu thanks for the PR. Have you had cases where handling termination / truncation has made a huge difference in the performance?

Josh00-Lu:
> Hi @Josh00-Lu thanks for the PR. Have you had cases where handling termination / truncation has made a huge difference in the performance?

Yes. Especially for mujoco environments like `swimmer`, where `truncation` events are encountered more frequently than `termination`. When `termination` is predominant, this problem is not as obvious (but explicit handling does not negatively affect performance). I think this improvement is intuitive, helping to use GAE for more reasonable Temporal Credit Assignment in PPO, as it explicitly recognizes the difference between termination and truncation. Admittedly, scaled extensive experiments may be needed to verify this.

gmarkkula:
+1 on seeing cases where it makes a big difference. CleanRL consistently performed much worse than SB3 on one of my environments, but when I added the bootstrapping from terminal observation to CleanRL it achieved the same performance as SB3. See [this comment](https://github.com/vwxyzjn/cleanrl/issues/198#issuecomment-2047350319) in another thread on this topic.

varadVaidya:
+1 on massive improvements using the PR's changes. I have a custom quadrotor improvement, that works well with SB3, but implementation of cleanRL gave poor performances, including catastrophic forgetting of some progress on the best seeds (the seeds which gave best performance on SB3). Using the PR's correct handling of the truncation, i have matching performance wrt SB3

Josh00-Lu:
> +1 on massive improvements using the PR's changes. I have a custom quadrotor improvement, that works well with SB3, but implementation of cleanRL gave poor performances, including catastrophic forgetting of some progress on the best seeds (the seeds which gave best performance on SB3). Using the PR's correct handling of the truncation, i have matching performance wrt SB3

Thank you. It's great to see that! 😄

sdpkjc:
Thank you for this PR! I had previously noticed the issue and made a simple modification to the CleanRL PPO by referencing the fix done in https://github.com/DLR-RM/stable-baselines3/pull/658. Here is my code:

https://gist.github.com/sdpkjc/71cbba3bd439d754ef9c026c31658ecd#file-ppo_continuous_action-py-L431-L437

It only requires adding a few lines to the existing code.

I found that this modification offers limited performance improvements in the default Mujoco environments, but it yields benefits in scenarios that frequently involve truncation.

I also noticed that Costa @vwxyzjn seems to have been interested in this issue previously. In his blog post https://iclr-blog-track.github.io/2022/03/25/ppo-implementation-details/, he mentioned this problem with PPO. However, he opted not to fix it in his implementation for the sake of fidelity to the original algorithm.

image

Whether or not to address this issue depends on whether CleanRL prioritizes higher performance or a more faithful implementation of the original algorithm. This is a topic worthy of our discussion. As far as I know, sb3 has addressed this issue.

Perhaps we could consider adding an argument to allow users to choose whether to handle this correctly.

Josh00-Lu:
> Thank you for this PR! I had previously noticed the issue and made a simple modification to the CleanRL PPO by referencing the fix done in [DLR-RM/stable-baselines3#658](https://github.com/DLR-RM/stable-baselines3/pull/658). Here is my code: https://gist.github.com/sdpkjc/71cbba3bd439d754ef9c026c31658ecd#file-ppo_continuous_action-py-L431-L437
> 
> It only requires adding a few lines to the existing code.
> 
> I found that this modification offers limited performance improvements in the default Mujoco environments, but it yields benefits in scenarios that frequently involve truncation.
> 
> I also noticed that Costa @vwxyzjn seems to have been interested in this issue previously. In his blog post https://iclr-blog-track.github.io/2022/03/25/ppo-implementation-details/, he mentioned this problem with PPO. However, he opted not to fix it in his implementation for the sake of fidelity to the original algorithm.
> 
> image
> Whether or not to address this issue depends on whether CleanRL prioritizes higher performance or a more faithful implementation of the original algorithm. This is a topic worthy of our discussion. As far as I know, sb3 has addressed this issue.
> 
> Perhaps we could consider adding an argument to allow users to choose whether to handle this correctly.

Thank you very much for your suggestions. I have carefully read your code section, but there might be some differences in implementation. There is a small issue with your approach, where the reward before truncation is replaced by `reward` + `value`. This might not be a good practice for a couple of reasons: 1) Many reward shaping techniques involve modifying the value rather than simply the rewards, 2) The same batch of PPO trajectories is used multiple times to update both the policy and value networks. See https://gist.github.com/sdpkjc/71cbba3bd439d754ef9c026c31658ecd#file-ppo_continuous_action-py-L500 
and https://github.com/vwxyzjn/cleanrl/blob/8cbca61360ef98660f149e3d76762350ce613323/cleanrl/ppo_continuous_action.py#L259
Therefore, directly modifying the original reward and setting it as a fixed value might not be advisable (my personal view). Consequently, this pr only uses environmental states or flag to handle truncations, and have led to certain performance improvements in relevant tasks.

I have added an argument for my logic, and I will update the PR soon.

vwxyzjn:
Thanks all for the discussion. I had opted out to implement the correct behavior also because at some point I did a comprehensive eval and it did not seem to matter. There are also issues with efficient implementation. The current gymnasium's design is highly inefficient that we need to do extra forward passes whenever there is the terminal states.


 https://github.com/sail-sg/envpool/issues/194#issuecomment-1317171243

Josh00-Lu:
> Thank you for this PR! I had previously noticed the issue and made a simple modification to the CleanRL PPO by referencing the fix done in [DLR-RM/stable-baselines3#658](https://github.com/DLR-RM/stable-baselines3/pull/658). Here is my code: https://gist.github.com/sdpkjc/71cbba3bd439d754ef9c026c31658ecd#file-ppo_continuous_action-py-L431-L437
> 
> It only requires adding a few lines to the existing code.
> 
> I found that this modification offers limited performance improvements in the default Mujoco environments, but it yields benefits in scenarios that frequently involve truncation.
> 
> I also noticed that Costa @vwxyzjn seems to have been interested in this issue previously. In his blog post https://iclr-blog-track.github.io/2022/03/25/ppo-implementation-details/, he mentioned this problem with PPO. However, he opted not to fix it in his implementation for the sake of fidelity to the original algorithm.
> 
> image
> Whether or not to address this issue depends on whether CleanRL prioritizes higher performance or a more faithful implementation of the original algorithm. This is a topic worthy of our discussion. As far as I know, sb3 has addressed this issue.
> 
> Perhaps we could consider adding an argument to allow users to choose whether to handle this correctly.

Thank you. I have carefully read the codes you provided, but there might be some differences in implementation. There is a small issue with your approach: it involves replace the `reward` before the truncation with `reward`+`value`. While some reward shaping methods involve value shaping or recomputing (e.g., the recompute advantage tricks), and the same batches of PPO trajectories are used to update the value networks multiple times. Therefore, directly modifying the original reward and setting it as a fixed value might not be a good approach (my opinion). Consequently, this pr tries to use only environmental states and flags for truncation handling.

pseudo-rnd-thoughts:
Sorry for the late message, personally, I would advocate for a separate file with this implementation if the original isn't wanting to be changed. This would prevent the problem people report above that they would need to make the change themselves. 
Second, @vwxyzjn Gymnasium v1.0.0 is updated to the vector implementation to match EnvPool in autoreset style, i.e., no more `final_observation` therefore, most of CleanRL will need to be updated to reflect this change

Josh00-Lu:
> Sorry for the late message, personally, I would advocate for a separate file with this implementation if the original isn't wanting to be changed. This would prevent the problem people report above that they would need to make the change themselves. Second, @vwxyzjn Gymnasium v1.0.0 is updated to the vector implementation to match EnvPool in autoreset style, i.e., no more `final_observation` therefore, most of CleanRL will need to be updated to reflect this change. Meaning that the change suggested here will not cause the additional forward pass.
> 
> Therefore, my suggest would be to reopen this PR, include it in a new file, `ppo_truncted.py`?? and update to Gymnasium v1.0.0 within or after

I agree with your idea. I have re-opened this PR and included it in a new file called ppo_continuous_action_truncted.py. If other files in this repository are updated to Gymnasium v1.0.0, I will make the following adjustments to this file accordingly.

PR #456, opened 2024-04-15
Removing the unused noise_clip from sac_continuous_actions.py

jonas-eschmann:
Removing the unused `noise_clip` from `sac_continuous_actions.py` (probably a relic from TD3)


vwxyzjn (approved):
Thanks!

PR #472, opened 2024-07-17
Add Parallel Q-Networks algorithm (PQN)

roger-creus:
## Description
Adding PQN from [Simplifying Deep Temporal Difference Learning](https://arxiv.org/pdf/2407.04811)

I have implemented both `pqn.py` and `pqn_atari_envpool.py`. The results are promising for the Cartpole version. Check them out [here](https://wandb.ai/rogercreus/cleanRL/reports/CleanRL-PQN-Cartpole-v1--Vmlldzo4NzExNTkx). I am now running some debugging experiments for the Atari version.

Some details about the implementations:
- Both use envpool
- Hyperaprameters try to match the configs from the [official implementations](https://github.com/mttga/purejaxql/tree/main/purejaxql/config) but some are changed (epsilon-decay schedule matches the DQN implementation from cleanRL. I haven't checked the importance of the hyperparameter the defaults in CleanRL made more sense to me)
- For comparing `pqn.py` and `dqn.py` in cartpole I multiplied the rewards from the environment by 0.1 as done in the official implementation of PQN. performance increases for both algos.
- Using LayerNorm in the networks instead of allowing the user to select between Layer or Batch norm. Layer norm should work better. 
- Not giving the user the option to add BatchNorm to the inputs to the network (i.e. states) as in the official implementaiton.

Overall the implementation is similar to ppo with envpool (so very fast!) but with the sample-efficiency of Q-learning! Nice algorithm! :)

Let me know how to proceed! 

## Types of changes

- [ ] Bug fix
- [ ] New feature
- [x] New algorithm
- [ ] Documentation

## Checklist:

    - [x] I have added links to the original paper and related papers.

- [x] I have contacted @vwxyzjn to obtain access to the [openrlbenchmark W&B team](https://wandb.ai/openrlbenchmark).
- [x] I have used the [benchmark utility](/get-started/benchmark-utility/) to submit the tracked experiments to the [openrlbenchmark/cleanrl](https://wandb.ai/openrlbenchmark/cleanrl) W&B project, optionally with `--capture_video`.
- [x] I have performed RLops with `python -m openrlbenchmark.rlops`.
    - For new feature or bug fix:
    - For new algorithm:
    - [x] I have added the learning curves generated by the `python -m openrlbenchmark.rlops` utility to the documentation.
    - [x] I have added links to the tracked experiments in W&B, generated by `python -m openrlbenchmark.rlops ....your_args... --report`,  to the documentation.


sdpkjc:
Hey Roger, it's really cool to see you adding PQN to CleanRL! I've read the paper before, and I think your implementation is great. When it comes time to run benchmarks or add documentation, let's collaborate to see how we can best do it. Looking forward to seeing the completed PR! 🚀👍

sdpkjc:
> I think the code might be ready to be benchmarked. These are some results in Breakout. It seems to converge to 400 score in 10M which would match DQN. The official imlpementation reports 515 score after 400M steps. Should I be added to the `openrlbenchmark` W&B team?
> 
> image

I noticed that the epsilon greedy implementation in our current setup differs from the official one, where each agent independently performs epsilon greedy exploration, whereas in our implementation, all environments share a single random number. This might have an impact when running many environments in parallel. Of course, there could be other reasons for the performance differences too. Let's start by running some benchmark tests to see if the performance also falls short in other environments. Looking forward to working through this together!

https://github.com/mttga/purejaxql/blob/9878a74439593c5d0acc8e506fefc44daa230c51/purejaxql/pqn_atari.py#L312-L325

roger-creus:
Very nice catch! Let me try to set up the benchmark experiments :)

roger-creus:
Here are some [first results](https://api.wandb.ai/links/openrlbenchmark/qh5l0u8p)! 
I think they look pretty good but maybe in `BeamRider-v5` is falling a bit short. Let me double check the implementation and run some more experiments

vwxyzjn:
Been watching this from far, very cool work!!

pseudo-rnd-thoughts:
Nice job, your results show it takes 25 minutes for 10 million frames while the paper reports 200 million in an hour. 
Do you know why there are such significant differences in performance?

roger-creus:
Updated results [here](https://api.wandb.ai/links/openrlbenchmark/qh5l0u8p). I wonder how should I generate the comparison between DQN/PQN with the `rlops` function since I am using `envpool` and I am not being able to compare `pqn_atari_envpool.py` in `Breakout-v5` vs `dqn_atari.py` in `BreakoutNoFrameskip-v4` for instance. Should I make a version of PQN that doesn't use envpool?

@pseudo-rnd-thoughts It is probably because `jax.lax.scan`. I am not used to coding in jax, but after searching online, it seems pytorch does not have a function like `scan`...

sdpkjc:
Maybe try `torch.compile`?

roger-creus:
Hey! How do you think we should proceed? 

I believe that it will be hard to match the speed of the JAX-based original implementation in this torch implementation, but at least it provides a Q-learning alternative + envpool that matches CleanRL envpool PPO, which can already be very useful! :)

roger-creus:
I realized I was re-computing the values for each state in the rollouts when computing `Q(lambda)` returns. I have now used a `values` buffer (as used in the PPO implementations actually) and replaced the computations in the `Q(lambda)` process. Performance remains the same and the code is now approx **150% faster**. 

Also, I added `pqn_atari_lstm_envpool`! First results in the atari environments show the implementation is correct. Please double check! :)

Please let me know how we should continue!

pseudo-rnd-thoughts:
@roger-creus There is a larger issue of EnvPool with rollouts and computing the loss function, see https://github.com/vwxyzjn/cleanrl/issues/475

roger-creus:
@vwxyzjn Hey! Following up :) What more do you think should be done for this PR?

vwxyzjn:
Hey @roger-creus thanks for the ping and the nice RP! Could you add documentation like the other algorithms?

cleanrl/dqn.py
```diff
@@ -69,6 +69,8 @@ class Args:
     """timestep to start learning"""
     train_frequency: int = 10
     """the frequency of training"""
+    rew_scale: float = 0.1
```
vwxyzjn, line 72: Let's no add this to dqn.

cleanrl/pqn_atari_envpool.py
```diff
@@ -0,0 +1,284 @@
...
+    """the learning rate of the optimizer"""
+    num_envs: int = 128
+    """the number of parallel game environments"""
+    num_steps: int = 32
+    """the number of steps to run in each environment per policy rollout"""
+    anneal_lr: bool = True
+    """Toggle learning rate annealing for policy and value networks"""
+    gamma: float = 0.99
+    """the discount factor gamma"""
+    num_minibatches: int = 32
```
vwxyzjn, line 53: The number of minibatches seems quite high. Where is this coming from? When comparing to other parameters I noticed quite some difference as well.

image


It's also inconsistent with the pqn_atari_lstm_envpool
image

cleanrl/pqn_atari_envpool.py
```diff
@@ -0,0 +1,284 @@
...
+            nn.LayerNorm([64, 9, 9]),
+            nn.ReLU(),
+            nn.Conv2d(64, 64, 3, stride=1),
+            nn.LayerNorm([64, 7, 7]),
+            nn.ReLU(),
+            nn.Flatten(),
+            nn.Linear(3136, 512),
+            nn.LayerNorm(512),
+            nn.ReLU(),
+            nn.Linear(512, env.single_action_space.n),
```
vwxyzjn, line 128: Maybe do the `layer_init` to PQN as well like in PPO?

cleanrl/pqn_atari_lstm_envpool.py
```diff
@@ -0,0 +1,329 @@
...
+        infos["r"] = self.returned_episode_returns
+        infos["l"] = self.returned_episode_lengths
+        return (
+            observations,
+            rewards,
+            dones,
+            infos,
+        )
+
+
```
vwxyzjn, line 110: Let's remove non-meaningful lines of code differences like this extra new line:

image

cleanrl/pqn.py
```diff
@@ -0,0 +1,274 @@
+# docs and experiment results can be found at https://docs.cleanrl.dev/rl-algorithms/dqn/#dqnpy
+import os
+import random
+import time
+from collections import deque
+from dataclasses import dataclass
+
+import envpool
```
vwxyzjn, line 8: What do you think on making this more similar to `ppo.py` using gymnasium? Envpool doesn't work on windows for example (I think)

image

vwxyzjn (commented):
Added some comments on style. Overall we should try to reduce the lines of code differences against a reference file (e.g., pqn.py and ppo.py or ppo_atari_envpool.py and pqn_atari_envpool.py)

cleanrl/pqn_atari_envpool.py
roger-creus: Actually for now I will just leave it to `num_envs=128`, `num_steps=32`, `num_minibatches=32` and once this PR is merged I can make a new one to decide the best hyperparameters for both PQN and PPO (envpool versions).

cleanrl/pqn.py
roger-creus: Done! 👍

roger-creus:
Thanks for your review @vwxyzjn !

- `pqn.py` now uses gymnasium instead of envpool (I ran some experiments in CartPole and it continues to work).
- `pqn_atari_envpool.py` and `pqn_atari_envpool_lstm.py` now have the same rollout hyperparameters as PPO.
- [Validation experiments](https://api.wandb.ai/links/rogercreus/wza159vn) are successful -- they match PPO. See that PQN-LSTM experiments are now running.
- I added a documentation page but I am unsure how to add the plots/curves/tables for performance with RLOps. I tried but got an error, maybe because the logs are currently in my WandB account. 

Let me know how to proceed! :)

vwxyzjn:
Hi @roger-creus sorry the slow response. Lots of going on lately. Could you clone the openrlbenchmark repo and run the following? Namely:

```
git clone https://github.com/openrlbenchmark/openrlbenchmark.git
poetry install
python -m openrlbenchmark.rlops \
    --filters '?we=rogercreus&wpn=cleanRL&ceik=env_id&cen=exp_name&metric=charts/episodic_return' \
        'pqn_atari_envpool?tag=pr-472&cl=CleanRL PQN' \
    --env-ids Breakout-v5 SpaceInvaders-v5 BeamRider-v5 Pong-v5 MsPacman-v5 \
    --no-check-empty-runs \
    --pc.ncols 3 \
    --pc.ncols-legend 3 \
    --rliable \
    --rc.score_normalization_method maxmin \
    --rc.normalized_score_threshold 1.0 \
    --rc.sample_efficiency_plots \
    --rc.sample_efficiency_and_walltime_efficiency_method Median \
    --rc.performance_profile_plots  \
    --rc.aggregate_metrics_plots  \
    --rc.sample_efficiency_num_bootstrap_reps 10 \
    --rc.performance_profile_num_bootstrap_reps 10 \
    --rc.interval_estimates_num_bootstrap_reps 10 \
    --output-filename static/0compare \
    --scan-history
```

I can't do the plots because I don't have access to the cleanrl project. Could you make it public? Thanks.

image

vwxyzjn:
I tested it with a new install and openrlbenchmark seems to work fine.

image

roger-creus:
Thanks for the reply @vwxyzjn ! I have updated the documentation, added the plots, etc. I have checked the changes locally and I think it looks fine! All tests pass, etc.

This PR is pretty much ready in my opinion, but let me know if you would like some additional details or experiments! :)

roger-creus:
@vwxyzjn Just following up :) Let me know if you'd like anything else in this PR

vwxyzjn:
This is an amazing PR. Merging as is. Thanks for the great work @roger-creus!!! I am also sent you an invite as a collaborator for CleanRL. Feel free to merge the PR.

vwxyzjn (approved):

Issue #475, opened 2024-07-25
PPO Envpool doesn't account for episode change 


pseudo-rnd-thoughts:
## Problem Description
For PPO rollouts, it is critical to account for episode changes that occur when an environment autoresets.
For Gym (and Gymnasium < 1.0), the vector environment autoreset within the step and included in the actual final observation within the info with the observation returned for `done==True` being the next episode's observation. 
This means that the cleanrl gym based rollout's correctly account for this, as the next episode's observation is nullified by `done`. 

However, this strategy doesn't work for EnvPool (and most critically gymnasium == 1.0 alphas). 
These vector environments shift the next episode reset for the step after `done==True`. 
Meaning that on the current implementation within the rollout observations, we have `obs=[t_0, t_1, t_2, t_0, t_1]` and `done=[False, False, True, False]` for an episode of three observations and one observation of the next. 

In my head (I still need to confirm working example), we need to ignore the loss for the `done+1` if this makes sense, i.e., the obs/next obs for after an episode ends. 
I've been scrolling through the Gym and EnvPool and don't see any code that already does this (I might be wrong, apologies if I am).

## TL;DR
**Current Behaviour**: EnvPool PPO computes the loss between the last and first observations of new episodes (critical as Gymnasium 1.0 is shifting to an EnvPool style reset). 
**Expected Behavior**: The loss between the last and first observation should be zeroed / masked out

## Possible Solution
Include a mask on the loss / advantage for `done + 1` to ignore these values.

## Steps to Reproduce
To do, largely theoretically thinking on my side. I suspect this hasn't been realised as EnvPool is relatively niche and episodes are long enough that it is difficult to spot

sdpkjc:
I encountered this problem while using the EnvPool formal interface in my self-driving project. I decided to mask the done+1 samples, implementing the masking as the samples are extracted from the replay buffer to ensure the continuity of the environment iteration remains unaffected.

EdanToledo:
Hey, i thought I'd also chime in here. I realised this difference and i simply made a wrapper to achieve the same auto-reset style as Gym API. My wrapper is [here](https://github.com/EdanToledo/Stoix/blob/main/stoix/wrappers/envpool.py) if that helps.

PR #494, opened 2025-01-13
Fix: PQN needs to mask the return computation if next state is done

roger-creus:
## Description
In PQN, the `nextnonterminal` is currently used to mask the next value prediction:

```
returns[t] = rewards[t] + args.gamma * (
    args.q_lambda * returns[t + 1] + (1 - args.q_lambda) * next_value * nextnonterminal
)
```

but it should also mask the next return, which can be fixed as:

```
returns[t] = rewards[t] + args.gamma * (
    args.q_lambda * returns[t + 1] + (1 - args.q_lambda) * next_value
) * nextnonterminal
```

This PR fixes this issue for all `pqn.py`, `pqn_atari_envpool.py` and `pqn_atari_envpool_lstm.py`

## Types of changes

- [x] Bug fix
- [ ] New feature
- [ ] New algorithm
- [x] Documentation

## Checklist:

    - [x] I have added links to the original paper and related papers.

- [x] I have contacted @vwxyzjn to obtain access to the [openrlbenchmark W&B team](https://wandb.ai/openrlbenchmark).
- [x] I have used the [benchmark utility](/get-started/benchmark-utility/) to submit the tracked experiments to the [openrlbenchmark/cleanrl](https://wandb.ai/openrlbenchmark/cleanrl) W&B project, optionally with `--capture_video`.
- [x] I have performed RLops with `python -m openrlbenchmark.rlops`.
    - For new feature or bug fix:
        - [x] I have used the RLops utility to understand the performance impact of the changes and confirmed there is no regression.
    - For new algorithm:
    - [x] I have added the learning curves generated by the `python -m openrlbenchmark.rlops` utility to the documentation.
    - [x] I have added links to the tracked experiments in W&B, generated by `python -m openrlbenchmark.rlops ....your_args... --report`,  to the documentation.


pseudo-rnd-thoughts (commented):
Do you not need to zero the error for nextnonterminal rather than the return? Or are they equivalent?

roger-creus:
> Do you not need to zero the error for nextnonterminal rather than the return? Or are they equivalent?

@pseudo-rnd-thoughts No, I don't think so. In a terminal state, your **TD target return** is the episode's last reward (which is corrected in this PR), and you need to regress that (hence your TD error might not be zero). In a non-terminal state, your TD target return (with Q lambda) is the reward at that state plus some recursive factor of the returns from the future states. You always need to regress the TD targets so the loss shouldn't be zeroed out.

pseudo-rnd-thoughts:
Sorry I think I might be thinking of a different problem that I thought this was fixing. 

See https://github.com/vwxyzjn/cleanrl/issues/475 for more detail.

In short, in a rollout, you will have the end of one episode and the first of the next episode.
The problem with the original PQN and I think your implementation is that we don't account for this and still compute the error between the last and first observations of neighbouring episodes. 

I hope that makes sense and to solve this issue requires zero outs this particular observation to prevent learning from this error

roger-creus:
@pseudo-rnd-thoughts  This is a known limitation in all current CleanRL+Envpool implementations. The problem arises because the observation returned by EnvPool when `done=True` is the final observation of the episode. CleanRL's current implementation incorrectly masks the error between the penultimate and the final observations of each episode while it should mask the error between the final and the first observations of the next episode. For example, here’s the current observation accompanying a `done=True` signal:

doneOld_13064

I fixed this error by modifying the `RecordEpisodeStatistics` wrapper in the envpool scripts to account for the auto-resetting differences between envpool and gymnasium, [as done here](https://github.com/EdanToledo/Stoix/blob/0dcf4102cd6b1fa5ea4224c1f6b72480ee4343a9/stoix/wrappers/envpool.py#L76)

Like this:

```
def step(self, action):
        observations, rewards, dones, infos = super().step(action)
        
        # Fix envpool auto-resetting
        env_ids_to_reset = np.where(dones)[0]
        if len(env_ids_to_reset) > 0:
            (
                reset_obs,
                _,
                _,
                _,
            ) = self.env.step(np.zeros_like(action), env_ids_to_reset)
            observations[env_ids_to_reset] = reset_obs
        
        self.episode_returns += infos["reward"]
        self.episode_lengths += 1
        self.returned_episode_returns[:] = self.episode_returns
        self.returned_episode_lengths[:] = self.episode_lengths
        self.episode_returns *= 1 - infos["terminated"]
        self.episode_lengths *= 1 - infos["terminated"]
        infos["r"] = self.returned_episode_returns
        infos["l"] = self.returned_episode_lengths
        return (
            observations,
            rewards,
            dones,
            infos,
        )
```

In this way, the errors that will be masked are between the second last observation of the episode (now taken as the true last) and the first observation of the next episode. With these changes, now the observation accompanying the `done=True` signal is the following:

doneFixed_11392

I can include this in the PQN implementations or do it for all the envpool implementations in CleanRL to close #475. Let me know.

Tagging @vwxyzjn  as well :)

pseudo-rnd-thoughts:
While I agree that this is an option, personally I dislike it. 
I would prefer to keep all the observations as we can terminate and truncate correctly with all the data and then conduct masking of the loss function to solve the issue of autoreset. 
I know this adds some complexity but it should be possible.

roger-creus:
> While I agree that this is an option, personally I dislike it. 
> I would prefer to keep all the observations as we can terminate and truncate correctly with all the data and then conduct masking of the loss function to solve the issue of autoreset. 
> I know this adds some complexity but it should be possible. 

Understood! Still, this PR fixs a different issue regarding the return computation which needs to be fixed as well

pseudo-rnd-thoughts:
>  Still, this PR fixes a different issue regarding the return computation which needs to be fixed as well

total, sorry for derailing this pr

vwxyzjn:
yeah the auto reset stuff is a bit tricky, also I did a benchmark a while ago and did not find too much performance impact, so overall it feels not worth it to fix it.

The PR looks good to me @roger-creus. Would this be a performance-impacting change? if so please re-run the benchmark exps to make sure no significant regression. Otherwise feel free to merge.

roger-creus:
@vwxyzjn I ran the regression check experiments and while I accidentally removed the previous results from wandb, I was still able to put together a regression check [available here](https://docs.google.com/presentation/d/1x8RJEc_SEDWXnZPM23Nk8PCxl5H9Q7W0lhnUucFuw8Q/edit?usp=sharing).

As you will see, this fix yields significant performance gains across several environments. I have updated the plots in the documentation, the commands to obtain the benchmark results and the wandb space. This PR is now finished and I will merge once you double check the results! :)

PR #495, opened 2025-01-14
[Non-performance-impacting update] Use Pytorch DDP in ppo_atari_multigpu

realAsma:
[Non-performance-impacting update] Use Pytorch DDP in ppo_atari_multigpu

## Description
`cleanrl/ppo_atari_multigpu.py` does not use any data parallel wrappers and performs gradient synchronization explicitly.
Since Pytorch distributed parallel wrappers have become more ubiquitous, we could use them instead and improve the code readability.

## Types of changes

- [ ] Bug fix
- [x] New feature
- [ ] New algorithm
- [ ] Documentation

## Checklist:

    - For new feature or bug fix:
    - For new algorithm:

### DIFF
--- a/cleanrl/ppo_atari_multigpu.py
+++ b/cleanrl/ppo_atari_multigpu.py
@@ -234,7 +234,9 @@ E.g., `torchrun --standalone --nnodes=1 --nproc_per_node=2 ppo_atari_multigpu.py
 
     agent = Agent(envs).to(device)
     if args.world_size > 1:
+        # DDP syncs gradients (after each backward step), weights are sync'd at DDP initialization
         agent = DDP(agent)
+
     optimizer = optim.Adam(agent.parameters(), lr=args.learning_rate, eps=1e-5)
 
     # ALGO Logic: Storage setup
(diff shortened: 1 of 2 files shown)


vwxyzjn:
Hi @realAsma, thanks for the PR. I am a bit more inclined to keep things as is because it showcases the lower-level implementation of DDP, so maybe more informative to certain users. That said I like your implementation as well. Maybe you could add in the docs a link to this PR, saying you could have implemented multi-gpu this way?

realAsma:
> Hi @realAsma, thanks for the PR. I am a bit more inclined to keep things as is because it showcases the lower-level implementation of DDP, so maybe more informative to certain users. That said I like your implementation as well. Maybe you could add in the docs a link to this PR, saying you could have implemented multi-gpu this way?

@vwxyzjn I made a minor update to this PR to update the docs. Now I like the implementation in this PR. In the prior implementation, weight initialization across processes are handled by setting torch seed to be the same across process before model initialization. After model initialization, torch seeds are set back to the unique seed across processes. Additionally, There is that boiler-plate code for gradient synchronization.

The prior implementation is more educational overall. However, do we intend to teach distributed training concepts as well? 
My thinking is that DDP example showcases the PPO with multi-GPU/multi-process workflow better by being more succinct. 

I support whichever works for you :). I will create a different PR which adds a link to this once in the docs if you wish so. Thanks a lot for taking your time to review this PR. Let me know what you think.

vwxyzjn:
I still like the existing implementation for being more educational, but I agree the DDP implementation is more standard. If you are motivated, we can also just add another file for it like `ppo_atari_ddp.py` or `ppo_atari_multigpu_ddp.py`, and you can run the benchmark experiments to compare with the existing implementation.

Alternative feel free to create a different PR adding the link to this one in the docs. Either way is fine with me and up to you.

PR #502, opened 2025-03-02
Upgrade gymnasium to 1.1.0

sdpkjc:
## Types of changes

- [ ] Bug fix
- [ ] New feature
- [ ] New algorithm
- [ ] Documentation

## Checklist:

    - For new feature or bug fix:
    - For new algorithm:


pyproject.toml
```diff
@@ -12,13 +12,13 @@ license="MIT"
...
-python = ">=3.8,<3.11"
+python = ">=3.9,<3.11"
 tensorboard = "^2.10.0"
 wandb = "^0.13.11"
 gym = "0.23.1"
 torch = ">=1.12.1"
-stable-baselines3 = "2.0.0"
-gymnasium = ">=0.28.1"
+stable-baselines3 = ">=2.4.0"
+gymnasium = ">=1.0.0"
```
pseudo-rnd-thoughts, line 21: Can this be specified as v1.1.0, if sb3 is the limitation then I think see if I can update it

pseudo-rnd-thoughts (commented):
Thanks for starting this @sdpkjc 

Ale-py should be updated to v0.10.1

And the auto reset mode of the vector environment should be updated

pseudo-rnd-thoughts:
If I remember correctly, SB3 is used for the replay buffer and the atari wrappers. IMO, those features can probably be shifted in `cleanrl_utils` directly

cleanrl_utils/same_model_vector_env.py
```diff
@@ -0,0 +1,34 @@
+from copy import deepcopy
+
+import gymnasium as gym
+import numpy as np
+
+
+# Only for gymnasium v1.0.0
+class SameModelSyncVectorEnv(gym.vector.SyncVectorEnv):
```
pseudo-rnd-thoughts, line 8: This should probably be called `SameStepModeSyncVectorEnv` or we just shift to gymnasium v1.1.0

pyproject.toml
```diff
@@ -12,13 +12,13 @@ license="MIT"
 readme = "README.md"
 
 [tool.poetry.dependencies]
-python = ">=3.8,<3.11"
+python = ">=3.9,<3.11"
```
pseudo-rnd-thoughts, line 15: What is limiting us increasing this?

pyproject.toml
```diff
@@ -12,13 +12,13 @@ license="MIT"
...
 tensorboard = "^2.10.0"
 wandb = "^0.13.11"
 gym = "0.23.1"
 torch = ">=1.12.1"
-stable-baselines3 = "2.0.0"
-gymnasium = ">=0.28.1"
+stable-baselines3 = "^2.4.0"
```
pseudo-rnd-thoughts, line 20: I suspect we will have a new release of sb3 with support for gymnasium v1.1.0 as no changes seem to be required on their end (https://github.com/DLR-RM/stable-baselines3/pull/2095)

cleanrl_utils/same_model_vector_env.py
pseudo-rnd-thoughts: https://farama.org/Vector-Autoreset-Mode

MarcusBinderDTU:
Hey, with the updates in gymnasium 1.1 would it not be easier to simply use the 'Same-Step Mode' or am I missing something? 

Does it have to do with the support for the other wrappers that are only supported in the 'Next step' mode?

pseudo-rnd-thoughts:
@MarcusBinderDTU it is more about minimising implementation changes. 
The old implementation used same step, therefore, as this is a module update, we are trying to minimise code changes. 
Moving to next step (which for the  PPO implementations could be beneficial) would be a separate or

MarcusBinderDTU:
> @MarcusBinderDTU it is more about minimising implementation changes. The old implementation used same step, therefore, as this is a module update, we are trying to minimise code changes. Moving to next step (which for the PPO implementations could be beneficial) would be a separate or

Thanks for the fast reply! 

I agree, but I dont understand why not going directly to gymnasium 1.1 and then using `autoreset_mode=gym.vector.AutoresetMode.SAME_STEP` for the environment creation? 

Would that not be the easiest way of doing it?

sdpkjc:
Thanks for your suggestion. However, since the currently released version of sb3 depends on gymnasium < 1.1, we can’t upgrade to 1.1 directly. Once https://github.com/vwxyzjn/cleanrl/pull/505 is merged, we'll remove the sb3 dependency and then update to gymnasium 1.1, which will allow us to use `autoreset_mode=gym.vector.AutoresetMode.SAME_STEP` for environment creation. So this PR is on hold until https://github.com/vwxyzjn/cleanrl/pull/505 goes in.

MarcusBinderDTU:
Ahh, now I see! Thanks for clarifying it, that makes sense :)

cleanrl/dqn_atari.py
jugheadjones10, line 1: For this line:
real_next_obs[idx] = infos["final_observation"][idx]
I think "final_observation" should be "final_obs".
Tried running and "final_observation" gives a key error but not "final_obs".

varadVaidya:
Do you think its a good idea to let the users know that the logging for gymnasium >= 1.0 is different until the PR is merged? I did face some puzzling errors, and took more time than it should have, since the change in reward logging is not mentioned anywhere prominently in the changelogs for gymnasium.

pseudo-rnd-thoughts:
Closing in favour of https://github.com/vwxyzjn/cleanrl/pull/516

PR #538, opened 2026-02-11
fix: use single_action_space in DDPG for vectorized envs

Mr-Neutr0n:
## Bug
DDPG uses `envs.action_space` instead of `envs.single_action_space` when constructing networks, which gives incorrect shapes in vectorized environments.

## Fix
Changed to `single_action_space` to get the correct per-environment action dimensions.

### DIFF
--- a/cleanrl/ddpg_continuous_action.py
+++ b/cleanrl/ddpg_continuous_action.py
@@ -102,10 +102,10 @@ class Actor(nn.Module):
         self.fc_mu = nn.Linear(256, np.prod(env.single_action_space.shape))
         # action rescaling
         self.register_buffer(
-            "action_scale", torch.tensor((env.action_space.high - env.action_space.low) / 2.0, dtype=torch.float32)
+            "action_scale", torch.tensor((env.single_action_space.high - env.single_action_space.low) / 2.0, dtype=torch.float32)
         )
         self.register_buffer(
-            "action_bias", torch.tensor((env.action_space.high + env.action_space.low) / 2.0, dtype=torch.float32)
+            "action_bias", torch.tensor((env.single_action_space.high + env.single_action_space.low) / 2.0, dtype=torch.float32)
         )
 
     def forward(self, x):


Mr-Neutr0n:
Quick note on the CI failures here: they're not caused by this PR.

The 13 failing jobs all fail in the same step: `uv pip install ".[pytest, atari, jax]"`. The `[jax]` extra in `pyproject.toml` (line 52–58) pins `jax==0.4.8`, `jaxlib==0.4.7`, `flax==0.6.8`, `optax==0.1.4`, `chex==0.1.5` — and `jaxlib==0.4.7` is no longer available on the index that uv resolves against:

```
× No solution found when resolving dependencies:
  ╰─▶ Because there is no version of jaxlib==0.4.7 and cleanrl[jax]==2.0.0b1
      depends on jaxlib==0.4.7
```

The actual code change here (1 line: `env.action_space` → `env.single_action_space` in DDPG) is correct, and `tests/test_atari.py` passes in every job that gets to it before the `[jax]` install step fails.

Other open PRs against `master` fail in the same way (e.g. `srygaard/ppo-atari-envpool-zerocopy`, run 26189758574) — so the broken `[jax]` pin is a project-wide pre-existing issue, not a regression from this change. Probably worth a separate PR to bump the jax extra to current versions (jax 0.6+ ships with matching jaxlib and Flax 0.10+).

Happy to send that follow-up if useful — but figured I'd flag the diagnosis first so this PR doesn't get stuck on something unrelated.

Mr-Neutr0n:
I opened a focused follow-up PR for the broken JAX extra pins here: #554.

That PR updates the stale `jax`/`jaxlib`/`flax`/`optax`/`chex` versions to a set that resolves cleanly on Python 3.10 and includes a small local validation (`uv pip install -e '.[pytest,jax]'` plus `tests/test_jax_compute_gae.py`).

Keeping it separate so this DDPG fix can be evaluated on its own merits instead of staying blocked on the unrelated dependency issue.

Mr-Neutr0n:
Closing this one to tidy up my open pull requests.

It's been open around five months with no review activity, and I'd rather withdraw it than leave it sitting in your queue. Nothing needed from you, and no hard feelings at all.

If the fix is still wanted, this can be reopened, or I'm happy to redo it properly against current main. Apologies for the noise.

PR #569, opened 2026-08-12
docs: point the tensorboard command at the repo root where runs/ is created

### LINKED ISSUE #568: Misleading TensorBoard Command
In the README,
```bash
# open another terminal and enter `cd cleanrl/cleanrl`
tensorboard --logdir runs
```
implies that the runs directory is in cleanrl/cleanrl. In fact, runs is in cleanrl.

Powfu-zwx:
Fixes #568.

The quickstart runs `uv run python cleanrl/ppo.py` from the repository root, and `SummaryWriter(f"runs/{run_name}")` creates `runs/` relative to the working directory — the repository root, not `cleanrl/cleanrl`. The comment now tells readers to open the second terminal in the same directory instead of cd-ing into the package folder.


Powfu-zwx:
Closing this while I clean up my forks. The fix itself is a one-liner - pointing the TensorBoard --logdir at the repository root - so feel free to apply it directly if useful.

